PR #7, opened 2017-12-13
Add a Travis check for code formatting.

g-easy:
- The helper script can also be used to autoformat the code.

- Format the code.

- Make bazel under Travis less chatty.


g-easy:
Here's an example of what it looks like when it fails: https://travis-ci.org/g-easy/opencensus-cpp/builds/315714854#L573

opencensus/trace/internal/trace_config.cc
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "opencensus/trace/internal/trace_config_impl.h"
 #include "opencensus/trace/trace_config.h"
+#include "opencensus/trace/internal/trace_config_impl.h"
```
bogdandrutu: This seems to be a bug in the clang-format. Can you report?

.travis.yml
```diff
@@ -9,5 +9,6 @@ before_install:
   - ./bazel-0.8.1-installer-linux-x86_64.sh --user
 
 script:
-  - bazel build -k //...
-  - bazel test -k //...
+  - bazel build --noshow_progress -k //...
+  - bazel test --noshow_progress -k //...
+  - ./format.sh
```
bogdandrutu: I think travis should just verify that is well formatted. travis cannot change the code correct?

g-easy: Right. Travis doesn't make a commit. format.sh uses clang-format in place, reports which files were modified, and returns exitcode 1 for Travis.

opencensus/trace/internal/trace_config.cc
g-easy: 1. I think the problem here is we should have had a blank line after the _impl include.

2. Travis has clang-format v3.9.0 (Sep 2016), which is different from what I have on my workstation, and produces different results. My workaround was to make format.sh look for clang-format-3.9. Devs on Debian systems can simply "apt install clang-format-3.9"

g-easy: 1. On second thoughts, isn't this the correct order? trace_config.cc includes trace_config.h first.

bogdandrutu: Then shouldn't be an empty line between the main ".h" include and the rest of the includes?

g-easy: Done. :)

g-easy:
Travis silently upgraded clang-format to v5.0.0 on their end and now the check is failing. :^)

Vizerai:
LGTM

PR #9, opened 2017-12-20
Make trace/ Exporter API consistent with the stats/ one.

g-easy:
Also:
- Add sampler_test.
- Replace Span's default constructor with BlankSpan().
- Add BlankSpan test.
- Initialize threads after the rest of the exporter, to avoid racing.


Vizerai:
LGTM

PR #11, opened 2018-01-17
Use the same copts as ABSL.

g-easy:
In particular, -Wno-sign-compare as Google style does not use unsigned
integers.

### DIFF
--- a/opencensus/stats/examples/BUILD
+++ b/opencensus/stats/examples/BUILD
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load("//opencensus:copts.bzl", "TEST_COPTS")
+
 licenses(["notice"])  # Apache 2.0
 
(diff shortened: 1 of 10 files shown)


g-easy:
What do you folks think? The motivation for this change is to silence the "warning: comparison between signed and unsigned integer expressions [-Wsign-compare]"

If we decide to go ahead with this, I need to add copts=... to every target in every BUILD file. This is unfortunate, but it's how ABSL already does it.

opencensus/stats/examples/BUILD
```diff
@@ -1,6 +1,6 @@
 # OpenCensus C++ Stats library: Example usage.
 #
-# Copyright 2017, OpenCensus Authors
+# Copyright 2017, 2018, OpenCensus Authors
```
isturdy: Are we supposed to be updating these dates? Abseil is not.

PR #16, opened 2018-01-22
Add a benchmark for stats, presently covering aggregation/aggregation

g-easy:
window combinations and varying numbers of views.


Vizerai:
LGTM

PR #20, opened 2018-02-06
Fix export intervals.

isturdy:
Delay creation of exporter threads until a handler is registered, and do not include the time taken by the export in the export interval.


g-easy:
Please fix //opencensus/stats:stats_exporter_test

isturdy:
Fixed (passed 500 repetitions).

PR #28, opened 2018-02-15
Add a benchmark for batched recording (as in the gRPC plugin).

isturdy:


g-easy:
Travis: //opencensus/plugins:stats_plugin_end2end_test                          TIMEOUT

opencensus/stats/internal/stats_manager_benchmark.cc
```diff
@@ -110,11 +110,75 @@ BENCHMARK_TEMPLATE2(BM_Record, DistributionAggregation, CumulativeWindow)
...
+            .add_column(tag_key_1)
+            .add_column(tag_key_2);
+    views.push_back(absl::make_unique<View>(distribution_descriptor));
+  }
+
+  std::vector<std::string> tag_values(10);
+  for (int i = 0; i < 10; ++i) {
+    tag_values[i] = absl::StrCat("value", i);
+  }
+  int iteration = 0;
```
g-easy: Optional: wrap everything above L162 in state.PauseTiming()/ResumeTiming()

isturdy: The benchmark library does not start timing until the first call to the state loop--as I understand PauseTiming()/ResumeTiming() are only needed if you are doing housekeeping inside the loop.

isturdy:
Merging since the travis failure is irrelevant (and already being worked on).

Issue #29, opened 2018-02-16
stats_plugin_end2end_test is flaky


isturdy:
There is a race condition between checking stats in the main thread and recording them in the server thread, which is leading to Travis failures. (2% flaky on my workstation, possibly more in Travis.)

PR #30, opened 2018-02-16
Wait and retry checking server stats in the stats_plugin_end2end_test.

isturdy:
Server stats are updated by the server thread after the server responds, creating a possible race condition. This commit completely avoids race failures, although still leaves minor (~0.1%) flakiness from RPC failures (see issue #29).


isturdy:
It seems that the lost-RPC rate on travis is prohibitive--abandoning this attempt in favor of simply blacklisting the test until I work out what is wrong.

PR #38, opened 2018-02-20
Change BucketBoundaries' internal constructor to take a vector by value.

### LINKED ISSUE #21: Use std::move reduce BucketBoundaries copy overhead
I found BucketBoundaries when construction is through internal will make a copy, here to move construction will be better

```
BucketBoundaries(absl::Span<const double> lower_boundaries)
      : lower_boundaries_(lower_boundaries.begin(), lower_boundaries.end()) {}

  // The lower bound of each bucket, excluding the underflow bucket but
  // including the overflow bucket.
  std::vector<double> lower_boundaries_;

Call: 
  std::vector<double> boundaries(num_finite_buckets + 1);
  for (int i = 0; i <= num_finite_buckets; ++i) {
   ..........
  }
  return BucketBoundaries(boundaries);
```

replace

```
  BucketBoundaries(std::vector<double>&& lower_boundaries)
  std::vector<double> boundaries(num_finite_buckets + 1);
  for (int i = 0; i <= num_finite_buckets; ++i) {
   ..........
  }
  return BucketBoundaries(std::move(boundaries));
```

isturdy:
This better accords with conventions (https://github.com/census-instrumentation/opencensus-cpp/blob/master/opencensus/doc/conventions.md#classes-that-take-an-object-and-store-it) and avoids a copy for the Linear and Exponential constructors.

Closes #21

### DIFF
--- a/opencensus/stats/internal/bucket_boundaries.cc
+++ b/opencensus/stats/internal/bucket_boundaries.cc
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <iostream>
+#include <vector>
 
 #include "absl/base/macros.h"
 #include "absl/strings/str_cat.h"
@@ -39,7 +40,7 @@ BucketBoundaries BucketBoundaries::Linear(int num_finite_buckets, double offset,
     boundaries[i] = boundary;
     boundary += width;
   }
-  return BucketBoundaries(boundaries);
+  return BucketBoundaries(std::move(boundaries));
 }
 
 // static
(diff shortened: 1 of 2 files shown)


opencensus/stats/internal/bucket_boundaries.cc
```diff
@@ -64,8 +64,7 @@ BucketBoundaries BucketBoundaries::Explicit(
     ABSL_ASSERT(0);
     return BucketBoundaries({});
   }
-  return BucketBoundaries(
-      absl::Span<const double>(boundaries.begin(), boundaries.size()));
+  return BucketBoundaries(std::vector<double>(boundaries));
```
bogdandrutu: You can change the Explicit to take a vector by value as well.

Issue #22, opened 2018-02-09
Something's wrong with AttributeValueRef or VS2017


gjasny:
Hello,

With Visual Studio 2017 (15.5.6) I noticed a failing test even in debug mode:
```
TEST(AttributeValueTest, StringValue) {
  AttributeValue attr(absl::StrCat("hello", " ", "world"));
  EXPECT_EQ(AttributeValue::Type::kString, attr.type());
  EXPECT_EQ("hello world", attr.string_value());
}
```

The temporary `AttributeValueRef` that was used to construct the `AttributeValue` contained a `string_view` which content pointer pointer into glibberish memory. When I changed the test to the following:

```
TEST(AttributeValueTest, StringValue) {
  std::string value = absl::StrCat("hello", " ", "world");
  AttributeValue attr(value);
  EXPECT_EQ(AttributeValue::Type::kString, attr.type());
  EXPECT_EQ("hello world", attr.string_value());
}
```

I noticed that before the `string_view` was created a temporary `std::string` got created, too. But instead of being kept alive until the `AttributeValue` was constructed, it gets deleted right before entry of `AttributeValueRef` ctor. Starting from that the string view points into free'd memory.

I thing the following AttributeValueRef code is to blame (or Visual Studio 2017)
```
  template <typename T, typename std::enable_if<std::is_constructible<
                            absl::string_view, T>::value>::type* = nullptr>
  AttributeValueRef(T string_value)
      : string_value_(absl::string_view(string_value)), type_(Type::kString) {}
```

If I change the signature to `AttributeValueRef(const T& string_value)`, the error is gone. I'm no C++ language lawyer but AFAIR temporaries are best passed as `const&`.

g-easy:
I wonder why we didn't run into this with clang+ASAN. We'll have to investigate.

isturdy:
As I read the standard, the temporary should last until the completion of the enclosing expression--constructing an AttributeValue from a temporary string should be fine because the temporary should live until copied during the AttributeValue construction, but constructing a non-temporary AttributeValueRef from a temporary std::string results in a dangling pointer. There are some exceptions where temporary constructor arguments get destroyed when the constructor returns, but I do not think any apply here.

As a sidenote, AttributeValue(const AttributeValueRef& ref) should probably take its argument by value--the whole point of AttributeValueRef is to be trivially copyable. (Likewise, there is little point in AttributeValueRef being movable.) But I do not think that is at fault here.

[...]

PR #41, opened 2018-02-22
Don't pass std::string by value to AttributeValueRef()

g-easy:
This causes an accidental copy of the string, and (with VS2017) frees
that string before it's used:
https://github.com/census-instrumentation/opencensus-cpp/issues/22

### DIFF
--- a/opencensus/trace/internal/span_test.cc
+++ b/opencensus/trace/internal/span_test.cc
@@ -16,6 +16,7 @@
 
 #include <cstdint>
 
+#include "absl/strings/str_cat.h"
 #include "gtest/gtest.h"
 #include "opencensus/trace/attribute_value_ref.h"
 #include "opencensus/trace/exporter/attribute_value.h"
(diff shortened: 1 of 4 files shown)


bogdandrutu:
Please add benchmarks for the following cases:
* AttributeValueRef("literal long string to prove that we do not copy");
* AttributeValueRef(my_long_string); where my_long_string has 100>chars

Please also add the following cases in the span_test:
* AddAttributes where value is constructed with StrCat;
* AddAttributes where key is constructed with StrCat;
* Maybe replicate these tests for all functions that accept AttributesRef and AddAttribute

g-easy:
$ bazel run -c opt //opencensus/trace:attribute_value_ref_benchmark

Before:

Benchmark                        Time           CPU Iterations
---------------------------------------------------------------
BM_ConstructFromLiteral          2 ns          2 ns  424733380
BM_ConstructFromString         205 ns        205 ns    3380308

After:

Benchmark                        Time           CPU Iterations
---------------------------------------------------------------
BM_ConstructFromLiteral          2 ns          2 ns  398711129
BM_ConstructFromString           2 ns          2 ns  404455940

:)

opencensus/trace/internal/span_test.cc
```diff
@@ -174,6 +175,18 @@ TEST(SpanTest, FullSpanTest) {
...
 
+  // String (as opposed to literal).
+  std::string val = "value";
+  val += "6";
+  span.AddAttribute("key6", val);
+
+  // Test that StrCat's output outlives the AddAttribute call.
+  span.AddAttribute(absl::StrCat("ke", "y7"), absl::StrCat("val", "ue7"));
+  span.AddAttributes(
+      {{absl::StrCat("key", "10"), "value10"},
```
bogdandrutu: Please also test that the values are there 👍

g-easy: Done. Thanks!

PR #49, opened 2018-02-23
Add bazel.rc and presubmit script.

g-easy:
The bazel.rc file has configs for address sanitizer (--config=asan) and
ubsan.

The presubmit script is a heavyweight version of tools/ci.sh which
builds and tests everything, including opt mode and sanitizers.

### DIFF
--- a/tools/presubmit.sh
+++ b/tools/presubmit.sh
@@ -44,7 +44,6 @@ buildables="-- $(bazel query -k --noshow_progress "kind('^cc', //...)")"
 
 tests="-- $(bazel query -k --noshow_progress \
   "kind(test, //...) \
-   except attr('tags', 'noci', //...) \
    except attr('tags', 'manual', //...)")"
 
 run bazel build $buildables
--- a/tools/presubmit.sh
+++ b/tools/presubmit.sh
@@ -1,4 +1,4 @@
-#! /usr/bin/env bash
+#!/usr/bin/env bash
 #
 # Copyright 2018, OpenCensus Authors
 #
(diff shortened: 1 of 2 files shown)


g-easy:
This takes ~8 minutes to run, and changing --config blows away the bazel cache, but it's easier and more thorough than having to "bazel build $(bazel query etc" manually. :)

g-easy:
With `cc_output_directory_tag` the cache is retained. A no-op run of presubmit.sh takes 4 sec.

tools/presubmit.sh
```diff
@@ -0,0 +1,63 @@
...
+  fi
+}
+
+t0="$(date +%s)"
+
+buildables="-- $(bazel query -k --noshow_progress "kind('^cc', //...)")"
+
+tests="-- $(bazel query -k --noshow_progress \
+  "kind(test, //...) \
+   except attr('tags', 'noci', //...) \
```
isturdy: If this is intended for running locally rather than on Travis I think this exemption should be removed--the main use for "noci" to this point is the stats_plugin_end2end_test which fails on Travis due to network issues. (Things that do not even run reliably locally, such as the Stackdriver stats e2e test, should be marked "manual" so that "bazel test opencensus/..." and the like work during development.)

g-easy: Good point. Done.

g-easy:
Travis build failure is due to a bug in ci.sh when no buildable targets are affected, fixed in https://github.com/census-instrumentation/opencensus-cpp/pull/55

PR #50, opened 2018-02-23
Add a Prometheus exporter

### LINKED ISSUE #44: Add a Prometheus exporter
We want to integrate support for exporting stats to Prometheus as well as Stackdriver. Prometheus uses a pull model--clients expose stats on an HTTP endpoint that the server scrapes--which will require an extension to the StatsExporter API. For exposing stats, I see two sensible options:
- Integrate with the existing Prometheus client library (https://github.com/jupp0r/prometheus-cpp).
- Expose a function returning formatted exports as a string, leaving it to application developers to expose it via HTTP.

The second minimizes our dependencies, but duplicates formatting code from the existing client library and makes it difficult for applications using the Prometheus client directly to integrate the two--they would have to implement their own Exposer combining the two sources of metrics. I prefer the first--if we use the existing client library and make the PrometheusExporter a custom Collectable, users can use the default Exposer for out-of-the-box export (supporting combining Opencensus and native Prometheus metrics), while still having the option of writing their own handler (using the client's formatters) to avoid running the Prometheus client's default server.

isturdy:
Closes #44.

### DIFF
--- a/opencensus/exporters/stats/prometheus/internal/prometheus_exporter.cc
+++ b/opencensus/exporters/stats/prometheus/internal/prometheus_exporter.cc
@@ -1,0 +1,40 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "opencensus/exporters/stats/prometheus/prometheus_exporter.h"
+
+#include <utility>
+#include <vector>
+
+#include "metrics.pb.h"
+#include "opencensus/exporters/stats/prometheus/internal/prometheus_utils.h"
+#include "opencensus/stats/stats.h"
+
+namespace opencensus {
+namespace exporters {
+namespace stats {
+
+std::vector<io::prometheus::client::MetricFamily>
+PrometheusExporter::Collect() {
+  const auto data = opencensus::stats::StatsExporter::GetViewData();
+  std::vector<io::prometheus::client::MetricFamily> output(data.size());
+  for (int i = 0; i < data.size(); ++i) {
+    SetMetricFamily(data[i].first, data[i].second, &output[i]);
+  }
+  return output;
+}
+
+}  // namespace stats
+}  // namespace exporters
+}  // namespace opencensus
--- a/opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
+++ b/opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
@@ -1,0 +1,128 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "opencensus/exporters/stats/prometheus/internal/prometheus_utils.h"
+
+#include <cctype>
+#include <cstdint>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "absl/strings/str_cat.h"
+#include "absl/strings/string_view.h"
+#include "absl/time/time.h"
+#include "metrics.pb.h"
+#include "opencensus/stats/stats.h"
+
+namespace opencensus {
+namespace exporters {
+namespace stats {
+
+namespace {
+
+// Replaces non-alphanumeric characters with underscores to satisfy
+// Prometheus's name requirements.
+std::string SanitizeName(absl::string_view name) {
+  std::string sanitized(name);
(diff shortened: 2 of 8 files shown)


opencensus/exporters/stats/prometheus/internal/prometheus_exporter.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "opencensus/exporters/stats/prometheus/internal/prometheus_utils.h"
+#include "opencensus/stats/stats.h"
+
+namespace opencensus {
+namespace exporters {
+namespace stats {
+
+std::vector<io::prometheus::client::MetricFamily>
+PrometheusExporter::Collect() {
+  const auto& data = opencensus::stats::StatsExporter::GetViewData();
```
g-easy: Why reference?

opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
```diff
@@ -0,0 +1,129 @@
...
+#include "opencensus/stats/stats.h"
+
+namespace opencensus {
+namespace exporters {
+namespace stats {
+
+namespace {
+
+// Replaces non-alphanumeric characters with underscores to satisfy
+// Stackdriver's name requirements.
```
g-easy: Prometheus's?

opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
```diff
@@ -0,0 +1,129 @@
...
+  histogram->set_sample_count(value.count());
+  histogram->set_sample_sum(value.count() * value.mean());
+
+  int64_t cumulative_count = 0;
+  for (int i = 0; i < value.bucket_boundaries().num_buckets(); ++i) {
+    cumulative_count += value.bucket_counts()[i];
+    auto* bucket = histogram->add_bucket();
+    bucket->set_cumulative_count(cumulative_count);
+    // We use lower boundaries plus an underflow bucket; Prometheus uses upper
+    // boundaries, including a +Inf boundary.
```
g-easy: `internal/prometheus_test_server.cc` does `BucketBoundaries::Explicit({0, 10})`

This code produces:
```
bucket{key1="v1",key2="",le="0.000000"} 0 1519637070515
bucket{key1="v1",key2="",le="10.000000"} 12 1519637070515
bucket{key1="v1",key2="",le="+Inf"} 18 1519637070515
```

Shouldn't it be just two buckets? Upper inclusive bounds 10 and +Inf?

opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
```diff
@@ -0,0 +1,129 @@
...
+}
+
+}  // namespace
+
+void SetMetricFamily(const opencensus::stats::ViewDescriptor& descriptor,
+                     const opencensus::stats::ViewData& data,
+                     io::prometheus::client::MetricFamily* metric_family) {
+  // TODO(sturdy): convert common units into base units (e.g. ms->s).
+  metric_family->set_name(
+      absl::StrCat(SanitizeName(descriptor.name()), "_",
```
g-easy: Probably faster to Sanitize(StrCat( than StrCat(Sanitize(

opencensus/exporters/stats/prometheus/internal/prometheus_utils_test.cc
```diff
@@ -0,0 +1,211 @@
...
+}
+
+TEST(StackdriverUtilsTest, MakeTimeSeriesDistributionDouble) {
+  const auto measure = opencensus::stats::MeasureRegistry::RegisterDouble(
+      "measure_distribution_double", "units", "");
+  const std::string view_name = "test_descriptor";
+  const std::string tag_key_1 = "foo";
+  const std::string tag_key_2 = "bar";
+  const auto bucket_boundaries =
+      opencensus::stats::BucketBoundaries::Explicit({0});
```
g-easy: Please test multiple buckets. :)

g-easy:
Optional: merge "Add an interface for pull exporters." separately from this PR.

opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
isturdy: Internally, we make three buckets out of `BucketBoundaries::Explicit({0, 10})`--an underflow bucket for `[-Inf, 0)`, and buckets `[0, 10)` and `[10, +Inf]`--for Prometheus we report the upper boundaries of those buckets. (Prometheus's rules prevent dropping the underflow bucket--we cannot drop its samples from the sum/mean, and since the bucket counts must equal the overall count we cannot omit any of the buckets without attributing its contents to a different bucket.)

opencensus/exporters/stats/prometheus/internal/prometheus_exporter.cc
isturdy: Changed--no good reason, given that GetViewData returns a value.

opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
isturdy: Yes.

opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
isturdy: True.

opencensus/exporters/stats/prometheus/internal/prometheus_utils_test.cc
isturdy: Done.

isturdy:
Split the first two commits (changes in opencensus/stats and opencensus/exporters/stats/stackdriver) to #59.

opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
g-easy: That makes sense but wasn't obvious to me.

Is this documented somewhere and I missed it? If not, please document it in a later PR. :)

isturdy: There is a brief comment in prometheus_utils.cc. I should write an md for each of the exporters documenting data model translations in greater depth.

PR #51, opened 2018-02-23
Add a simple gRPC benchmark.

isturdy:
This is likely not a very good benchmark (in particular, it somewhat unrealistically introduces client/server contention when recording stats) but provides a starting point for profiling.


bogdandrutu:
Can we also have a setup to run the official grpc loadtests/benchmarks with the plugin installed?

isturdy:
I would like to, but at the moment I cannot convince them to build on a clean clone, let alone with Census (and I do not see a clean way to inject Census without just copying and modifying their benchmark suite).

PR #54, opened 2018-02-24
Adding support for propagating context and retrieving span from context.

Vizerai:
Moving GenerateClientContext to a later callback so that it has access to the populated census_context from grpc.  Also adding in method for user to retrieve span from the census_context which gets stored in the grpc server_context and client_context.


isturdy:
This breaks //opencensus/plugins/grpc:stats_plugin_end2end_test. I can try to debug when I get in Monday.

Vizerai:
The test should work correctly now.

opencensus/plugins/grpc/internal/filter.cc
```diff
@@ -26,9 +26,19 @@ void GenerateServerContext(absl::string_view tracing, absl::string_view stats,
...
 
-void GenerateClientContext(absl::string_view method, CensusContext *context) {
-  // TODO: Change this so that it can be set to root or non-root span.
-  new (context) CensusContext(method);
+void GenerateClientContext(absl::string_view method, CensusContext *ctxt,
+                           CensusContext *parent_ctxt) {
+  if (parent_ctxt != nullptr) {
+    trace::SpanContext span_ctxt = parent_ctxt->Context();
+    trace::Span span = parent_ctxt->Span();
+    if (span_ctxt.span_id().IsValid() && span_ctxt.trace_id().IsValid()) {
```
g-easy: if (span_ctxt.IsValid())

opencensus/plugins/grpc/internal/filter.cc
```diff
@@ -26,9 +26,19 @@ void GenerateServerContext(absl::string_view tracing, absl::string_view stats,
...
+      new (ctxt) CensusContext(method, &span);
```
g-easy:
```
      new (method, parent);
      return;
    }
  }
  new (just method);
}
```

opencensus/plugins/grpc/internal/server_filter.cc
```diff
@@ -114,14 +115,12 @@ void CensusServerCallData::OnDoneRecvInitialMetadataCb(void *user_data,
...
-    GenerateServerContext(
-        absl::string_view(tracing_str, tracing_str_len),
-        absl::string_view(census_str, census_str_len), absl::string_view("", 0),
-        absl::string_view(method_str, calld->method_size_), &calld->context_);
-
-    absl::string_view method(method_str, calld->method_size_);
-    stats::Record({{RpcServerStartedCount(), 1}}, {{kMethodTagKey, method}});
+    GenerateServerContext(absl::string_view(tracing_str, tracing_str_len),
+                          absl::string_view(census_str, census_str_len),
+                          absl::string_view("", 0), calld->method_,
```
g-easy: /*primary_role=*/"", calld->method_,

g-easy:
I don't understand why tools/ci.sh is returning an error code, it doesn't look like there's any error from bazel.

g-easy:
"trap ERR" in ci.sh is causing problems, fixed in https://github.com/census-instrumentation/opencensus-cpp/pull/56 - merge that and this PR should pass Travis.

opencensus/plugins/grpc/internal/filter.cc
Vizerai: Done.

opencensus/plugins/grpc/internal/filter.cc
Vizerai: Done.

opencensus/plugins/grpc/internal/server_filter.cc
Vizerai: Done.

opencensus/plugins/grpc/internal/server_filter.cc
```diff
@@ -114,14 +115,12 @@ void CensusServerCallData::OnDoneRecvInitialMetadataCb(void *user_data,
...
-    GenerateServerContext(
-        absl::string_view(tracing_str, tracing_str_len),
-        absl::string_view(census_str, census_str_len), absl::string_view("", 0),
-        absl::string_view(method_str, calld->method_size_), &calld->context_);
-
-    absl::string_view method(method_str, calld->method_size_);
-    stats::Record({{RpcServerStartedCount(), 1}}, {{kMethodTagKey, method}});
+    GenerateServerContext(absl::string_view(tracing_str, tracing_str_len),
+                          absl::string_view(census_str, census_str_len),
+                          absl::string_view("", 0) /*primary_role=""*/,
```
g-easy: Drop the string_view and just: `/*primary_role="*/ ""`

opencensus/plugins/grpc/internal/server_filter.cc
```diff
@@ -203,10 +201,10 @@ void CensusServerCallData::Destroy(grpc_call_element *elem,
        {RpcServerRequestCount(), sent_message_count_},
        {RpcServerFinishedCount(), 1},
        {RpcServerResponseCount(), recv_message_count_}},
-      {{kMethodTagKey, absl::string_view(reinterpret_cast<char *>(
-                                             GRPC_SLICE_START_PTR(method_)),
-                                         method_size_)},
+      {{kMethodTagKey, method_},
        {kStatusTagKey, StatusCodeToString(final_info->final_status)}});
+  grpc_slice_unref_internal(path_);
+  context_.EndSpan();
```
g-easy: Good catch! :)

opencensus/plugins/grpc/internal/server_filter.cc
Vizerai: Done.

Issue #6, opened 2017-12-13
Add windows testing


bogdandrutu:
Probably use appveyor for that.

g-easy:
I've started work on this.

g-easy:
This is what I've got so far: https://github.com/g-easy/opencensus-cpp/tree/appveyor

It gets surprisingly far but build is broken on the proto targets with underscores that we filter out on Travis, and test is broken somewhere in protobuf. I'll try to resume work on this next week.

Example output: https://ci.appveyor.com/project/g-easy/opencensus-cpp

g-easy:
I'm running into `Cannot open compiler generated file: '': Invalid argument` - this is https://github.com/bazelbuild/bazel/issues/4149 (cl.exe can't handle long paths and gives a confusing error)

My workaround for now is:
```
mkdir c:\t
bazel --output_user_root=c:/t/ build //targets
```

isturdy:
Worked a bit on this in #81--presently blocked on boringssl selecting posix copts rather than windows. I am not certain what the cause is; they seem to be using the same setup that I successfully used for google benchmark.

isturdy:
Apparently gRPC pinned their boringssl dependency on a who-knows-how-old version with broken Windows builds under bazel. (Trying to fix with https://github.com/grpc/grpc/pull/14566)

bogdandrutu:
I think this is fixed. Please reopen if work is still left here.

g-easy:
We don't build everything and we don't test anything.

It looks like isturdy was working on (thanks!) getting the gRPC example / integration to build but ran into a grpc issue.

g-easy:
#292 replaced Appveyor Bazel build with a CMake one. I'd like to see Bazel come back, and also for everything to be built and tested like it is on Travis/Linux.

PR #58, opened 2018-02-26
Change lambda into IsBufferFull().

g-easy:
This is for compatibility with VS2017, which doesn't support
the unary plus trick for lambdas.


g-easy:
This is part of issue https://github.com/census-instrumentation/opencensus-cpp/issues/6

PR #61, opened 2018-02-27
Add registration interfaces to ViewDescriptor.

isturdy:
An API change for consideration: right now, registering a view for export uses a function in StatsExporter, and registering a view for on-task consumption uses the constructor of View. By moving those registration functions to ViewDescriptor, I think it becomes easier to find how to use it by following headers. (It also allows removing StatsExporter from the main user API--it will only be needed for unregistering views and writing exporters.)

WIP: needs implementation, removal/deprecation of competing functions.


isturdy:
Now I remember why I made View registration a constructor--I do not see a good way of returning a View without defining assignment, and as an RAII class I would rather not add that complexity.

Implementation/migration completed for StatsExporter::AddView() to ViewDescriptor::RegisterForExport().

PR #60, opened 2018-02-27
Add public GetSpanFromServerContext() API.

g-easy:

PR #63, opened 2018-02-27
Add gRPC example.

### LINKED ISSUE #12: gRPC example
Add an example of propagating trace information over a gRPC call.

g-easy:
Closes #12


opencensus/plugins/grpc/grpc_plugin.h
```diff
@@ -24,6 +29,9 @@ namespace opencensus {
 // defined below.
 void RegisterGrpcPlugin();
 
+// Returns the tracing Span for the current RPC.
+opencensus::trace::Span GetSpanFromServerContext(grpc::ServerContext* context);
```
g-easy: This comes from https://github.com/census-instrumentation/opencensus-cpp/pull/60

examples/grpc/hello_server.cc
```diff
@@ -0,0 +1,137 @@
...
+
+  // Expose a Prometheus endpoint.
+  prometheus::Exposer exposer("127.0.0.1:8080");
+  exposer.RegisterCollectable(exporter);
+
+  // Add views for RPC stats.
+  const auto server_request_bytes_descriptor =
+      opencensus::stats::ViewDescriptor()
+          .set_measure(opencensus::kRpcServerRequestBytesMeasureName)
+          .set_name("server_request_bytes")
```
isturdy: One nit--we have been encouraging uri-style names specifying aggregation and tags, e.g. opencensus.io/server_request_bytes-distribution-method.

g-easy: Per discussion in meeting: example.com/...

Also add comment.

g-easy: This code was removed in favor of RegisterExperimentalEtc. I'll send a separate PR for custom views.

opencensus/plugins/grpc/internal/grpc_plugin.cc
```diff
@@ -52,6 +55,11 @@ void RegisterGrpcPlugin() {
   RpcServerResponseCount();
 }
 
+opencensus::trace::Span GetSpanFromServerContext(grpc::ServerContext* context) {
```
Vizerai: We may want to expand this to allow for GetSpanFromClientContext as well.  This will require additional work though as the ClientContext initially just has a nullptr for the census_context.  We would need to setup functionality for populating it (this already exists for google3).  This works for now.

PR #65, opened 2018-02-27
Define canonical gRPC views.

isturdy:
Uses RegisterForExport() from #61.


opencensus/plugins/grpc/grpc_plugin.h
```diff
@@ -24,6 +25,12 @@ namespace opencensus {
 // defined below.
 void RegisterGrpcPlugin();
 
+// Registers the cumulative gRPC views so that they will be exported by any
```
g-easy: Please mark this API as experimental (even put Experimental in the function name) for now, until we sync up with the other languages.

isturdy: Done.

g-easy:
Thanks!

PR #69, opened 2018-02-28
Reenable the stats_plugin_end2end_test on ci.

isturdy:
Two changes:
- Switch to IPv4--for some reason this fixes the consistent timeouts, despite our Travis configuration nominally supporting IPv6 loopback.
- Add some short sleeps to give the server a chance to update stats before verifying.

### DIFF
--- a/opencensus/plugins/grpc/internal/stats_plugin_end2end_test.cc
+++ b/opencensus/plugins/grpc/internal/stats_plugin_end2end_test.cc
@@ -18,6 +18,8 @@
 
 #include "absl/strings/str_cat.h"
 #include "absl/strings/string_view.h"
+#include "absl/time/clock.h"
+#include "absl/time/time.h"
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 #include "include/grpc++/grpc++.h"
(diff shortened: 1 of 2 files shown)


bogdandrutu:
Is it possible to make the test deterministic (remove the sleeps)?

opencensus/plugins/grpc/internal/stats_plugin_end2end_test.cc
```diff
@@ -52,13 +54,13 @@ class StatsPluginEnd2EndTest : public ::testing::Test {
     // interface.
     ::grpc::ServerBuilder builder;
     int port;
-    builder.AddListeningPort("[::]:0", ::grpc::InsecureServerCredentials(),
+    builder.AddListeningPort("0.0.0.0:0", ::grpc::InsecureServerCredentials(),
```
g-easy: Please add a comment about IPv4 being less flaky on Travis.

isturdy:
To Bogdan's comment: the only way I have thought of is a loop checking stats and sleeping if they are not correct; the server thread is in a gRPC loop so there is not an easy way of having it mark completion.

PR #70, opened 2018-02-28
Check for span ended within End()

Vizerai:
Adding check for span ended in End() so it does not populate exporters with 2 copies of the span.

### DIFF
--- a/opencensus/trace/internal/span_impl.cc
+++ b/opencensus/trace/internal/span_impl.cc
@@ -136,7 +136,10 @@ void SpanImpl::SetStatus(exporter::Status&& status) {
 
 void SpanImpl::End() { EndWithTime(absl::Now()); }
 
-bool SpanImpl::HasEnded() { return has_ended_; }
+bool SpanImpl::HasEnded() {
+  absl::MutexLock l(&mu_);
+  return has_ended_;
+}
 
 void SpanImpl::EndWithLatencyForTesting(absl::Duration latency) {
   EndWithTime(start_time_ + latency);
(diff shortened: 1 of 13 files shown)


opencensus/trace/internal/span_impl.cc
```diff
@@ -136,6 +136,8 @@ void SpanImpl::SetStatus(exporter::Status&& status) {
 
 void SpanImpl::End() { EndWithTime(absl::Now()); }
 
+bool SpanImpl::HasEnded() { return has_ended_; }
```
g-easy: How does this work without a MutexLock? Looks like we're missing annotalysis on Travis.

g-easy:
Why is End() being called twice? This sounds like a bug.

Vizerai:
It isn't being called twice by default, but a user might easily make a mistake. Like in the case of the server span which is ended by the filter, the user might also end it themselves.

opencensus/trace/internal/span_impl.cc
Vizerai: I think technically we could get away without a lock here since it is just 1 byte being read, but let's put it in for good coding standards.

g-easy:
That's reasonable. I think we should add a warning for this case in debug mode.

opencensus/trace/internal/span_impl.h
```diff
@@ -87,6 +87,8 @@ class SpanImpl final {
 
   // Marks the end of the Span and sets its end_time_.
   void End() LOCKS_EXCLUDED(mu_);
+  // Returns true if the span has ended.
+  bool HasEnded() LOCKS_EXCLUDED(mu_);
```
g-easy: `bool HasEnded() const LOCKS_EXCLUDED(mu_);`

PR #72, opened 2018-02-28
Fixing bug in test.

Vizerai:


opencensus/trace/internal/local_span_store_test.cc
```diff
@@ -25,10 +25,10 @@ namespace trace {
 class SpanTestPeer {
  public:
   static void End(absl::Duration latency, Span* span) {
+    // Call Span::End to get into stores.
+    span->End();
```
g-easy: Won't this potentially store the span into the wrong latency bucket since latency hasn't been set yet?

Vizerai: Fixed now.

opencensus/trace/internal/span_impl.h
```diff
@@ -106,6 +106,7 @@ class SpanImpl final {
   friend class ::opencensus::trace::exporter::SpanExporterImpl;
   friend class ::opencensus::trace::SpanTestPeer;
 
+  void ExportSpanForTesting(const std::shared_ptr<SpanImpl>& span);
```
g-easy: 1. Let's submit this as-is to unbreak Travis.

2. This seems a strange place to put this function. It should be static since it's not using any members. And it's only used to add a shared_ptr wrapper around `*this`, which it doesn't have access to.

PR #83, opened 2018-03-02
Report status message when stats export fails.

g-easy:


g-easy:
I incorrectly used `error_details()` in a previous commit. It returns `a serialized google.rpc.Status proto` which apparently contains a NUL early enough to not display anything. :)

PR #89, opened 2018-03-06
Show view data before sleeping, move comments around.

g-easy:


examples/helloworld/hello_world.cc
```diff
@@ -76,31 +76,31 @@ int main(int argc, char **argv) {
   opencensus::stats::View view(video_size_view);
   video_size_view.RegisterForExport();
 
-  opencensus::trace::AlwaysSampler sampler;
+  static opencensus::trace::AlwaysSampler sampler;
```
Vizerai: Why static?

g-easy: Bogdan wants to instill this as a best practice because constructing a sampler is potentially expensive. So a single long-lived sampler should be used instead of e.g. creating one per Span.

I realize for a reduced example it doesn't make a real difference, but I'm trying to create good example code for people to copy and paste later. :)

Vizerai: Put a comment about this so the user knows.  I put a similar comment for making the measure local-static above.

g-easy: Good idea. Done.

PR #91, opened 2018-03-06
Add a readme for the Prometheus exporter.

isturdy:


opencensus/exporters/stats/prometheus/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+The Prometheus exporter exposes stats on an HTTP endpoint that is scraped by
+a Prometheus server. To install and run the Prometheus server, follow
+the instructions [here](https://prometheus.io/docs/introduction/first_steps/).
+
+### Register the exporter
+
+#### Using the Prometheus client library's exposer
+
+Register the Exposer on a selected port, and then register the Opencensus
```
g-easy: Register when? Before / after configuring views? Maybe hint to do it "in main" or "before starting the RPC server" or whatever the binary is doing?

PR #95, opened 2018-03-08
Add a TagSet class providing equality and hashing for key/value maps.

isturdy:
This is needed for the DeltaProducer--the ViewData maps include only the tag values in ViewDescriptor order, but since the DeltaProducer aggregates before view aggregation we need to keep the entire mapping.

### DIFF
--- a/opencensus/stats/internal/tag_set.cc
+++ b/opencensus/stats/internal/tag_set.cc
@@ -1,0 +1,70 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "opencensus/stats/tag_set.h"
+
+#include <functional>
+#include <initializer_list>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "absl/strings/string_view.h"
+
+namespace opencensus {
+namespace stats {
+
+TagSet::TagSet(
+    std::initializer_list<std::pair<absl::string_view, absl::string_view>>
+        tags) {
+  tags_.reserve(tags.size());
+  for (const auto& tag : tags) {
+    tags_.emplace_back(std::string(tag.first), std::string(tag.second));
+  }
+  Initialize();
+}
+
+TagSet::TagSet(std::vector<std::pair<std::string, std::string>> tags)
+    : tags_(std::move(tags)) {
+  Initialize();
+}
+
+void TagSet::Initialize() {
+  std::sort(tags_.begin(), tags_.end());
+
+  std::hash<std::string> hasher;
+  hash_ = 1;
+  for (const auto& tag : tags_) {
+    static const size_t kMul = static_cast<size_t>(0xdc3eb94af8ab4c93ULL);
+    hash_ *= kMul;
+    hash_ = ((hash_ << 19) |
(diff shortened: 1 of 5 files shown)


opencensus/stats/internal/tag_set_test.cc
```diff
@@ -0,0 +1,71 @@
...
+
+TEST(TagSetTest, EqualityRespectsKeyValuePairings) {
+  EXPECT_NE(TagSet({{"k1", "v1"}, {"k2", "v2"}}),
+            TagSet({{"k1", "v2"}, {"k2", "v1"}}));
+}
+
+TEST(TagSetTest, HashDisregardsOrder) {
+  TagSet ts1({{"k1", "v1"}, {"k2", "v2"}});
+  TagSet ts2({{"k2", "v2"}, {"k1", "v1"}});
+  EXPECT_EQ(TagSet::Hash()(ts1), TagSet::Hash()(ts2));
```
g-easy: This isn't testing the hash, this is testing that both tagsets are sorted in Initialize()

opencensus/stats/tag_set.h
```diff
@@ -0,0 +1,66 @@
...
+// between uses where possible.
+// TagSet is immutable.
+class TagSet final {
+ public:
+  // Both constructors are not explicit so that Record({}, {{"k", "v"}}) works.
+  // This constructor is needed because even though we copy to a vector
+  // internally because c++ cannot deduce the conversion needed.
+  TagSet(std::initializer_list<std::pair<absl::string_view, absl::string_view>>
+             tags);
+  // This constructor is needed so that callers can dynamically construct
```
g-easy: How about a single constructor that takes an absl::Span?

opencensus/stats/internal/tag_set.cc
```diff
@@ -0,0 +1,70 @@
...
+             (hash_ >> (std::numeric_limits<size_t>::digits - 19))) +
+            hasher(tag.first);
+    hash_ *= kMul;
+    hash_ = ((hash_ << 19) |
+             (hash_ >> (std::numeric_limits<size_t>::digits - 19))) +
+            hasher(tag.second);
+  }
+}
+
+std::size_t TagSet::Hash::operator()(const TagSet& tag_set) const {
```
g-easy: Is this functor is so that we can have an unordered_map<TagSet>? Should there be a test for that?

opencensus/stats/internal/tag_set_test.cc
isturdy: Given the implementation--I want it to catch anything silly (such as hashing before sorting).

opencensus/stats/tag_set.h
isturdy: That cannot autoconvert from e.g. `{{"key", "value"}}`--the compiler can convert the string literal to a string_view or the initialize list to a span, but not both.

opencensus/stats/internal/tag_set_test.cc
g-easy: That's reasonable.

opencensus/stats/tag_set.h
g-easy: Sorry, I didn't realize.

isturdy: No problem--absl::Span is what I initially wrote too. :)

opencensus/stats/internal/tag_set.cc
isturdy: Added one.

opencensus/stats/internal/tag_set.cc
```diff
@@ -0,0 +1,70 @@
...
+  Initialize();
+}
+
+void TagSet::Initialize() {
+  std::sort(tags_.begin(), tags_.end());
+
+  std::hash<std::string> hasher;
+  hash_ = 1;
+  for (const auto& tag : tags_) {
+    static const size_t kMul = static_cast<size_t>(0xdc3eb94af8ab4c93ULL);
```
g-easy: Let's leave this as-is for now.

But in a followup PR, please factor this against common/internal/string_vector_hash.h. I'd like to have a single place where this hash function (and its magic number) lives. :)

isturdy: Will do.

PR #96, opened 2018-03-09
Add the MeasureData class.

isturdy:
Another part of the DeltaProducer--this tracks all aggregations (including potentially multiple histograms) on a measure.

### DIFF
--- a/opencensus/stats/internal/measure_data.cc
+++ b/opencensus/stats/internal/measure_data.cc
@@ -1,0 +1,89 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "opencensus/stats/internal/measure_data.h"
+
+#include <algorithm>
+#include <cmath>
+#include <cstdint>
+#include <iostream>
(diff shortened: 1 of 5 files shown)


opencensus/stats/internal/measure_data.cc
```diff
@@ -0,0 +1,89 @@
...
+    histograms_.emplace_back(b.num_buckets());
+  }
+}
+
+void MeasureData::Add(double value) {
+  // Update using the method of provisional means.
+  ++count_;
+  ABSL_ASSERT(count_ > 0 && "Histogram count overflow.");
+  const double new_mean = mean_ + (value - mean_) / count_;
+  sum_of_squared_deviation_ =
```
g-easy: +=

opencensus/stats/internal/measure_data_test.cc
```diff
@@ -0,0 +1,143 @@
...
+
+  BucketBoundaries distribution_buckets = BucketBoundaries::Explicit({0});
+  Distribution distribution =
+      testing::TestUtils::MakeDistribution(&distribution_buckets);
+  EXPECT_DEBUG_DEATH(
+      {
+        data.AddToDistribution(&distribution);
+        EXPECT_THAT(distribution.bucket_counts(), ::testing::ElementsAre(1, 0));
+      },
+      "");
```
g-easy: If you ABSL_ASSERT(false && "Reason.") can you EXPECT("Reason")?

g-easy:
I've signed the CLA.

opencensus/stats/internal/measure_data_test.cc
isturdy: Yes. It even picks up the log on the previous line, so I do not need to duplicate the error message in the log and the assert.

opencensus/stats/internal/measure_data.cc
isturdy: Updating sum_of_squared_deviation requires both the new and old mean. I guess it is a bit clearer to save the old mean locally (formerly in the next line count_ is the new count and mean_ the old mean), but I do not think it saves anything.

PR #104, opened 2018-03-12
 Updating grpc plugin with OrphanablePtr.

Vizerai:
Updating grpc byte stream based on changes to grpc repo.


WORKSPACE
```diff
@@ -38,7 +38,7 @@ http_archive(
 # gRPC
 http_archive(
     name = "com_github_grpc_grpc",
-    urls = ["https://github.com/grpc/grpc/archive/master.tar.gz"],
+    urls = ["https://github.com/grpc/grpc/archive/master.zip"],
```
isturdy: Why change this? If the bazel bug was fixed, I think it should be its own commit.

Vizerai: It seems that travis does not like the zip file.  Reverting to the tar.gz file.

g-easy: Travis (dot yml) is installing an old version of bazel which doesn't have the symlink-in-zip-file fix.

That said, I don't think it's worth updating at this time, and not without discussing what minimum version of bazel we require.

PR #109, opened 2018-03-15
Change MBy to By units.

g-easy:
This reflects the default gRPC views which use bytes as units.


g-easy:
This is the C++ version of https://github.com/census-instrumentation/opencensus-java/pull/1058

Issue #112, opened 2018-03-19
examples/helloworld and website contain discrepancies on between bucket sizes and comments


odeke-em:
Hello there, I was just going over the examples here in order to instrument a C++ app and noticed that the examples have comments talking about
https://github.com/census-instrumentation/opencensus-cpp/blob/db9e2bb8b8b268c7cb36ef28dcf8f601b87877e2/examples/helloworld/helloworld.cc#L68
but the actual buckets aren't in those sizes
https://github.com/census-instrumentation/opencensus-cpp/blob/db9e2bb8b8b268c7cb36ef28dcf8f601b87877e2/examples/helloworld/helloworld.cc#L77
which are:
* 0
* 16 * 1MiB = 16MiB
* 256 * 1MiB = 256MiB

The website too seems to have discrepencies https://opencensus.io/cpp.html
<img width="648" alt="screen shot 2018-03-18 at 10 44 30 pm" src="https://user-images.githubusercontent.com/4898263/37579875-fe8d9a6c-2afd-11e8-98a5-52d6650fe2f6.png">

Please update these or if possible post them as plain values.

isturdy:
It seems this was an error in the Java quickstart that got copied over--fixing it now.

The example from the website is almost correct, but has the arguments wrong--it should be `Exponential(2, 1 << 8, 1 << 8)` (or `Exponential(2, 256, 256)`).

isturdy:
Fixed in all repositories I know it exists in. (Go's helloworld example does not have a similar comment.)

odeke-em:
Thank you @isturdy!

odeke-em:
Although, I still don't understand what 'Exponential(2, 256, 256)).` does.

isturdy:
It is explained in the header: https://github.com/census-instrumentation/opencensus-cpp/blob/master/opencensus/stats/bucket_boundaries.h#L40

(I believe we have a pending AI to more directly point people to the C++ headers from the website.)

odeke-em:
>(I believe we have a pending AI to more directly point people to the C++ headers from the website.)

Ahaha!

PR #113, opened 2018-03-19
Fix documentation of bucket boundaries in the helloworld example.

isturdy:
Fixes the opencensus-cpp part of #112.


Vizerai:
LGTM

PR #115, opened 2018-03-21
Zipkin exporter

Vizerai:
Zipkin exporter for opencensus-cpp
- Uses HTTP protocol with libcurl
- Uses JSON encoding with rapidjson


adriancole:
cc @llinder also

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+  writer->Key("name");
+  writer->String(std::string(span.name()));
+
+  writer->Key("traceId");
+  writer->String(span.context().trace_id().ToHex());
+
+  writer->Key("id");
+  writer->String(span.context().span_id().ToHex());
+
+  writer->Key("parentId");
```
adriancole: guard null (in other words avoid this field id absent)

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+  writer->String(span.parent_span_id().ToHex());
+
+  writer->Key("annotations");
```
adriancole: also avoid writing this collection if empty

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+    writer->StartObject();
+    writer->Key("timestamp");
+    writer->Int64(absl::ToUnixMicros(annotation.timestamp()));
+    writer->Key("value");
+    writer->String(annotation.event().DebugString());
+    writer->EndObject();
+  }
+  writer->EndArray(span.annotations().events().size());
+
+  writer->Key("binaryAnnotations");
```
adriancole: let's not do v1 format, or somehow permit v2 format. binaryAnnotations is archaic and replaced by "tags" which are always string -> string map  https://zipkin.io/zipkin-api/#/default/post_spans

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+        writer->String("I64");
+        break;
+      default:
+        break;
+    }
+    writer->EndObject();
+  }
+  writer->EndArray(span.attributes().size());
+
+  writer->Key("timestamp");
```
adriancole: in v2 format everything except trace id and span id are optional.. not sure which fields are optional in census, but be careful not to write zero values (like parentid = 0, timestamp= 0 etc)

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+namespace {
+
+constexpr const char *kZipkinLibname = "zipkin";
+constexpr const char *kZkipkinVersion = "2.0";
+
+void SerializeJson(rapidjson::Writer<rapidjson::StringBuffer> *writer,
+                   const ::opencensus::trace::exporter::SpanData &span) {
+  writer->StartObject();
+
+  writer->Key("name");
```
adriancole: at the very least we should (if v2 format) always write span.localEndpoint
if v1 format, span.localEndpoint is copied to every annotation and binary annotation

Either way, without this indexing on service or span name won't work.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+#include "absl/strings/str_cat.h"
+
+namespace opencensus {
+namespace exporters {
+namespace trace {
+
+namespace {
+
+constexpr const char *kZipkinLibname = "zipkin";
+constexpr const char *kZkipkinVersion = "2.0";
```
g-easy: Zkip -> Zip

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+#include <rapidjson/stringbuffer.h>
+#include "absl/strings/str_cat.h"
+
+namespace opencensus {
+namespace exporters {
+namespace trace {
+
+namespace {
+
+constexpr const char *kZipkinLibname = "zipkin";
```
g-easy: constexpr char kThing[]

(constexpr implies const, also * -> [])

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
g-easy: Timestamp and duration will always be present here, but parentId might not.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+  writer->String(span.parent_span_id().ToHex());
+
+  writer->Key("annotations");
+  writer->StartArray();
+  for (const auto &annotation : span.annotations().events()) {
+    writer->StartObject();
+    writer->Key("timestamp");
+    writer->Int64(absl::ToUnixMicros(annotation.timestamp()));
+    writer->Key("value");
+    writer->String(annotation.event().DebugString());
```
g-easy: DebugString is only for debugging. Please serialize the message and attributes explicitly.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+namespace opencensus {
+namespace exporters {
+namespace trace {
+
+namespace {
+
+constexpr const char *kZipkinLibname = "zipkin";
+constexpr const char *kZkipkinVersion = "2.0";
+
+void SerializeJson(rapidjson::Writer<rapidjson::StringBuffer> *writer,
```
g-easy: Output argument goes last.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
g-easy: Better yet, since these two constants are used together, just merge them here.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+    writer->EndObject();
+  }
+  writer->EndArray(span.attributes().size());
+
+  writer->Key("timestamp");
+  writer->Int64(absl::ToUnixMicros(span.start_time()));
+
+  writer->Key("duration");
+  writer->Int64(absl::ToInt64Microseconds(span.end_time() - span.start_time()));
+
```
g-easy: Links? MessageEvents?

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+  if (!curl) {
+    // Failed to create curl handle.
+    return;
+  }
+
+  const std::string mime_type = "application/json";
+  snprintf(content_type, sizeof(content_type), "Content-Type: %s",
+           mime_type.c_str());
+
+  headers = curl_slist_append(headers, content_type);
```
g-easy: Inline it all.

`headers = curl_slist_append(headers, "Content-Type: application/json");`

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+                                          &uploaded_bytes))) {
+        // Failed to get uploaded bytes.
+        error = curl_easy_strerror(res);
+      } else if (CURLE_OK !=
+                 (res = curl_easy_getinfo(curl, CURLINFO_SPEED_UPLOAD,
+                                          &upload_speed))) {
+        // Failed to get upload speed.
+        error = curl_easy_strerror(res);
+      } else {
+        // LOG(INFO) << "HTTP request finished, status " << status_code
```
g-easy: Remove debug code.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+        //           << ", uploaded " << uploaded_bytes << " bytes in "
+        //           << total_time << " seconds (" << (upload_speed / 1024)
+        //           << " KB/s)";
+      }
+    }
+  }
+
+  // TODO: Log error if one exists.
```
g-easy: Maybe: `if (!error.empty()) std::cerr << "curl error: " << error << "\n";` for now.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+  ::opencensus::trace::exporter::SpanExporter::RegisterHandler(
+      absl::WrapUnique<::opencensus::trace::exporter::SpanExporter::Handler>(
+          exporter));
+}
+
+void ZipkinExporter::Export(
+    const std::vector<::opencensus::trace::exporter::SpanData> &spans) {
+  if (!spans.empty()) {
+    std::string msg = message_codec_->Encode(spans);
+    trace_client_->SendMessage(msg, msg.size());
```
g-easy: http_client_

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+  }
+}
+
+void ZipkinExporter::ExportForTesting(
```
g-easy: IMO expose Encode() and unit-test that instead.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
```diff
@@ -0,0 +1,342 @@
...
+    const ZipkinExporterOptions &options,
+    const std::vector<::opencensus::trace::exporter::SpanData> &spans) {
+  if (!spans.empty()) {
+    // Create new exporter.
+    ZipkinExporter *exporter = new ZipkinExporter(options);
```
g-easy: Memory leak.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
```diff
@@ -0,0 +1,132 @@
+// Copyright 2018, OpenCensus Authors
```
g-easy: Please make this file look like e.g. opencensus/exporters/stats/stdout/stdout_exporter.h

Most of this file and the *.cc file should be moved to internal/

opencensus/exporters/trace/zipkin/zipkin_exporter.h
```diff
@@ -0,0 +1,132 @@
...
+class JsonCodec : public MessageCodec {
+ public:
+  const std::string name() const override { return "json"; }
+
+  std::string Encode(const std::vector<::opencensus::trace::exporter::SpanData>
+                         &spans) override;
+};
+
+// Pretty print JSON encoding
+class PrettyJsonCodec : public MessageCodec {
```
g-easy: I'm not sure we should support this.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
```diff
@@ -0,0 +1,132 @@
...
+  std::string Encode(const std::vector<::opencensus::trace::exporter::SpanData>
+                         &spans) override;
+};
+
+class ExportClient {
+ public:
+  virtual void SendMessage(const std::string &msg, size_t size) = 0;
+};
+
+struct ZipkinExporterOptions {
```
g-easy: A lot of this is duplicated as HttpClientOptions in the *.cc file.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Fixed.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
Vizerai: Removed HttpClient/HttpClientOptions and merged it with ZipkinExporter.  Also moved mostly everything to .cc file.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
Vizerai: Removed.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: I removed HttpClient and integrated it with ZipkinExporter.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: I didn't see an equivalent of those in the Zipkin format.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Put check around parentId.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: ExportForTesting directly calls encode now and prints the returned string.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: I updated it to record the localEndpoint now.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin.bzl
```diff
@@ -0,0 +1,192 @@
+# Copyright 2018, OpenCensus Authors
```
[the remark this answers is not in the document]
Vizerai: Renamed to curl.bzl.  I don't think it should go at the top level, so I left it in the zipkin folder for now.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,361 @@
...
+  writer->Key("name");
+  writer->String(std::string(span.name()));
+
+  writer->Key("traceId");
+  writer->String(span.context().trace_id().ToHex());
+
+  writer->Key("id");
+  writer->String(span.context().span_id().ToHex());
+
+  if (span.parent_span_id().IsValid()) {
```
adriancole: supernit : json looks neater if the parent is before the id

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,361 @@
...
+      return std::to_string(value.int_value());
+      break;
+  }
+  return "";
+}
+
+std::string SerializeAnnotation(
+    const ::opencensus::trace::exporter::Annotation &annotation) {
+  std::string annotation_str(annotation.description());
+  absl::StrAppend(&annotation_str, " (");
```
adriancole: not sure if others are doing this approach to serializing annotation attributes. At least if there are no attribute, I'd leave off the " ()" suffix as in zipkin annotations are equals search

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,361 @@
...
+      writer->Int64(absl::ToUnixMicros(annotation.timestamp()));
+      writer->Key("value");
+      writer->String(SerializeAnnotation(annotation.event()));
+      writer->EndObject();
+    }
+    writer->EndArray(span.annotations().events().size());
+  }
+
+  if (!span.attributes().empty()) {
+    writer->Key("tags");
```
adriancole: tags are not an array, they are a dict.

{ "foo": "bar" }

not
[ { "key": "foo", "value": "bar" } ]

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,361 @@
...
+      writer->String(AttributeValueToString(attribute.second));
+      writer->EndObject();
+    }
+    writer->EndArray(span.attributes().size());
+  }
+
+  writer->Key("timestamp");
+  writer->Int64(absl::ToUnixMicros(span.start_time()));
+
+  writer->Key("duration");
```
adriancole: unless end_time is guaranteed guard duration on it possibly not being present

Vizerai: It is always present in opencensus.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: Fixed.

WORKSPACE
```diff
@@ -104,3 +104,64 @@ load("@com_github_jupp0r_prometheus_cpp//:repositories.bzl",
...
+    name = "windows",
+    values = {"cpu": "x64_windows"},
+    visibility = [ "//visibility:private" ],
+)
+
+cc_library(
+    name = "curl",
+    srcs = glob([
+        "lib/*.c",
+        "lib/vauth/*.c",
```
g-easy: Should this be lib/**/*.c or are there dirs being excluded?

WORKSPACE
```diff
@@ -104,3 +104,64 @@ load("@com_github_jupp0r_prometheus_cpp//:repositories.bzl",
...
+        "lib/vtls/*.c",
+    ]),
+    hdrs = glob([
+        "include/curl/*.h",
+        "lib/*.h",
+        "lib/vtls/*.h",
+        "lib/vauth/*.h",
```
g-easy: Please sort.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,365 @@
...
+      return value.string_value();
+      break;
+    case ::opencensus::trace::AttributeValueRef::Type::kBool:
+      return value.bool_value() ? "true" : "false";
+      break;
+    case ::opencensus::trace::AttributeValueRef::Type::kInt:
+      return std::to_string(value.int_value());
+      break;
+  }
+  return "";
```
g-easy: Add something like ABSL_ASSERT(false && "Unknown AttributeValue type");

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,365 @@
...
+}
+
+std::string SerializeAnnotation(
+    const ::opencensus::trace::exporter::Annotation &annotation) {
```
g-easy: " &" -> "& " everywhere.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,365 @@
...
+
+  if (span.parent_span_id().IsValid()) {
+    writer->Key("parentId");
+    writer->String(span.parent_span_id().ToHex());
+  }
+
+  writer->Key("id");
+  writer->String(span.context().span_id().ToHex());
+
+  // Write endpoint.  Census does not support this by default.
```
g-easy: OpenCensus.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,365 @@
...
+  writer->Key("timestamp");
+  writer->Int64(absl::ToUnixMicros(span.start_time()));
+
+  writer->Key("duration");
+  writer->Int64(absl::ToInt64Microseconds(span.end_time() - span.start_time()));
+
+  writer->EndObject();
+}
+
+// JSON encoding
```
g-easy: "Returns a string which is all of the spans encoded as JSON."

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,365 @@
...
+
+  writer.StartArray();
+  for (const auto &span : spans) {
+    SerializeJson(span, service, &writer);
+  }
+  writer.EndArray();
+  return buffer.GetString();
+}
+
+std::string GetIpAddress(ZipkinExporterOptions::AddressFamily af_type) {
```
g-easy: I'm really wary of doing this here. Can we make the user supply it?

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,365 @@
...
+
+void CurlSendMessage(const uint8_t *data, const ZipkinExporterOptions &options,
+                     size_t size, CURL *curl, struct curl_slist *headers,
+                     char *err_msg) {
+  CURLcode res;
+
+  if (CURLE_OK !=
+      (res = curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, err_msg))) {
+    // Failed to set curl error buffer.
+    std::cerr << curl_easy_strerror(res);
```
g-easy: Just "return res" here and the rest of the function and have the caller handle logging.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,365 @@
...
+  char err_msg[CURL_ERROR_SIZE] = {0};
+  CURL *curl = curl_easy_init();
+
+  if (!curl) {
+    // Failed to create curl handle.
+    return;
+  }
+
+  headers = curl_slist_append(headers, "Content-Type: application/json");
+  headers = curl_slist_append(headers, "Expect:");
```
g-easy: Is this necessary? If so please add a comment to explain.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
```diff
@@ -0,0 +1,94 @@
...
+    // IP address will be local address of machine. It will be populated
+    // automatically.
+    std::string ip_address;
+  };
+};
+
+class ZipkinExporter
+    : public ::opencensus::trace::exporter::SpanExporter::Handler {
+ public:
+  void Export(const std::vector<::opencensus::trace::exporter::SpanData> &spans)
```
g-easy: Please make this look more like opencensus/exporters/stats/stdout/stdout_exporter.h

- Make ZipkinExporter not derive from Handler.
- Just define Register(options) in the public API, move Export() entirely to the *.cc file.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
```diff
@@ -0,0 +1,94 @@
...
+  friend class ZipkinExporterTestPeer;
+
+  static void ExportForTesting(
+      const std::vector<::opencensus::trace::exporter::SpanData> &spans);
+
+  ZipkinExporter(const ZipkinExporterOptions &options) : options_(options) {}
+  ~ZipkinExporter() {}
+
+  // Send HTTP message to zipkin endpoint using libcurl.
+  void SendMessage(const std::string &msg, size_t size) const;
```
g-easy: This shouldn't be here, this should only be in the *.cc file.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
```diff
@@ -0,0 +1,94 @@
...
+
+  ZipkinExporterOptions options_;
```
g-easy: Move both of these into the Handler class in the *.cc file.

opencensus/exporters/trace/zipkin/zipkin_exporter.cc
g-easy: Instead of printing the encoded string:
- Remove ExportForTesting entirely.
- Expose Encode() for testing.
- Construct a span, encode it, test that the encoded JSON equals what we expect.

g-easy:
@adriancole What, if anything, should we do about Attributes (span-level, not on annotations) and MessageEvents?

adriancole:
span.Attributes -> span.tags
span.MessageEvents[].description -> span.annotations

Not sure about the attributes on message events. they could be serialized like mentioned. @basvanbeek wdyt?

g-easy:
Sorry, I mis-spoke. I meant Links and MessageEvents.

@Vizerai has already taken care of Attributes (thanks!)

So Links and MessageEvents should both be converted to Zipkin annotations then?

adriancole:
links we have no mapping for.. seems they are not time based (from my
memory).. in such case I might make a "census.links" tag and json serialize
the list for now. We need to think it through more long term.

basvanbeek:
Heh you beat me to it @adriancole. Looking at the opencensus.io explanation of concepts it seems we do need to sharpen the definition of links and use cases for it. As it is easy for a novice consumer to be mistaken about it. 

Also I think being more explicit about which payloads are acceptable for MessageEvents is advisable including explanation of how we expect these to be mapped to potentially existing items in the trace exporters we support. In Zipkin we have our thrift definition to fall back on which holds all acceptable types of annotations, their use and their intended use case (see WIRE_SEND, WIRE_RECV, CLIENT_SEND_FRAGMENT, SERVER_RECV_FRAGMENT, etc.). Would be good to be explicit on mappings, add interop tests for these with the supported backends and thus make sure we get a consistent story between all the different language implementations for OC.

WORKSPACE
Vizerai: Done.

WORKSPACE
Vizerai: Done.

WORKSPACE
Vizerai: Done.

WORKSPACE
Vizerai: Done.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: Clang automatically formats it as " &".

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
Vizerai: Done.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
Vizerai: Moved to *.cc file.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: Added.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: I don't think these are strictly necessary.  Removed.

Vizerai:
@adriancole MessageEvents do not have a description string in opencensus.  They just have a type (SENT / RECEIVED), an id number, and a size.

Vizerai:
@adriancole is there an a simple zipkin server that can be run which can collect and parse the json data to  test if the exporter is working correctly?

adriancole:
You are all good on this, now, right? Using the default server from quick
start?

Vizerai:
Yes, it is working now.  Thanks!

Vizerai:
ping @googlebot

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: nvm it allows both styles.  Fixed now.

PR #123, opened 2018-03-22
Move measure registration to Measure from MeasureRegistry.

isturdy:
This completes moving away from direct interaction with the registries in the core stats workflows (after moving RegisterForExport from the StatsExporter to the ViewDescriptor). 

After this, the only use keeping MeasureRegistry in the public API is the possibility of measures shared between libraries with no common initialization (e.g. an rpc_latency measure shared between RPC libraries), for which we presently encourage looking up the measure and registering it if not found. I think we should consider recommending that such libraries either use individual measures or depend on a common measure definition package and make MeasureRegistry internal.

### DIFF
--- a/opencensus/stats/internal/measure.cc
+++ b/opencensus/stats/internal/measure.cc
@@ -14,11 +14,21 @@
 
 #include "opencensus/stats/measure.h"
 
+#include "absl/strings/string_view.h"
 #include "opencensus/stats/internal/measure_registry_impl.h"
 #include "opencensus/stats/measure_registry.h"
 
 namespace opencensus {
 namespace stats {
+
+// static
+template <typename MeasureT>
+Measure<MeasureT> Measure<MeasureT>::Register(absl::string_view name,
+                                              absl::string_view description,
+                                              absl::string_view units) {
+  return MeasureRegistryImpl::Get()->Register<MeasureT>(name, description,
+                                                        units);
+}
 
 template <typename MeasureT>
 const MeasureDescriptor& Measure<MeasureT>::GetDescriptor() const {
@@ -35,7 +45,7 @@ bool MeasureDouble::IsValid() const {
 }
 
 template <>
-bool MeasureInt::IsValid() const {
+bool MeasureInt64::IsValid() const {
   return MeasureRegistryImpl::IdValid(id_) &&
          MeasureRegistryImpl::IdToType(id_) == MeasureDescriptor::Type::kInt64;
 }
--- a/opencensus/stats/internal/stats_manager_test.cc
+++ b/opencensus/stats/internal/stats_manager_test.cc
@@ -16,7 +16,6 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 #include "opencensus/stats/measure.h"
-#include "opencensus/stats/measure_registry.h"
 #include "opencensus/stats/recording.h"
 #include "opencensus/stats/tag_key.h"
 #include "opencensus/stats/view.h"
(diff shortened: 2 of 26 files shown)


g-easy:
Does this make our API look more like the opencensus-java one?

opencensus/stats/internal/measure.cc
```diff
@@ -14,12 +14,22 @@
...
+#include "absl/strings/string_view.h"
 #include "opencensus/stats/internal/measure_registry_impl.h"
 #include "opencensus/stats/measure_registry.h"
 
 namespace opencensus {
 namespace stats {
 
+// static
+template <typename MeasureT>
+Measure<MeasureT> Measure<MeasureT>::Register(absl::string_view name,
```
g-easy: I'm sorry to be a pain but since this is already an API change, this would be a great time to go from (name,units,description) to (name,description,units) to match the argument order in opencensus-java and opencensus-go.

opencensus/stats/internal/stats_manager_test.cc
```diff
@@ -28,14 +28,14 @@ constexpr char kFirstMeasureId[] = "first_measure_name";
 constexpr char kSecondMeasureId[] = "second_measure_name";
 
 MeasureDouble FirstMeasure() {
-  static MeasureDouble measure = MeasureRegistry::RegisterDouble(
-      kFirstMeasureId, "ops", "Usage of resource 1.");
+  static MeasureDouble measure =
+      MeasureDouble::Register(kFirstMeasureId, "ops", "Usage of resource 1.");
```
g-easy: IMO use "1" instead of "ops" to match the standard.

opencensus/stats/measure.h
```diff
@@ -31,6 +31,30 @@ namespace stats {
...
+  // latency" shared between RPC libraries) check whether the measure is
+  // registered with MeasureRegistry::GetMeasure*ByName before registering it.
+  //
+  // 'name' should be a globally unique identifier. It is recommended that this
+  //   be in the format "<domain>/<path>", e.g. "example.com/client/foo_usage".
+  // 'units' are the units of recorded values. The recommended grammar is:
+  //     - Expression = Component { "." Component } {"/" Component }
+  //     - Component = [ PREFIX ] UNIT [ Annotation ] | Annotation | "1"
+  //     - Annotation = "{" NAME "}"
+  //   For example, string “MBy{transmitted}/ms” stands for megabytes per
```
g-easy: Cribbing from https://github.com/census-instrumentation/opencensus-specs/blob/afbb08857d94a7876ff69ec0b34e596da72f5c49/stats/gRPC.md

Latencies are measures in milliseconds, denoted "ms"
Sizes are measured in bytes, denoted "By"
Dimensionless values have unit "1"

opencensus/stats/measure_registry.h
```diff
@@ -22,37 +22,11 @@ namespace opencensus {
 namespace stats {
 
 // The MeasureRegistry keeps a record of all MeasureDescriptors registered,
-// providing functions for registering measures and querying their metadata by
-// name or handle.
+// providing functions for  querying their metadata by name or handle. Use
```
g-easy: Double space.

isturdy:
Java presently has `Measure{Double|Long}.create(String name, String description, String unit)`. Making this a static method of the measure classes is moving closer to Java; I think `Register` is more consistent with the rest of cpp than adopting `Create`.

opencensus/stats/measure_registry.h
isturdy: Done.

opencensus/stats/internal/measure.cc
isturdy: Done. (And MeasureInt->MeasureInt64, which I think is less important but still nice while we are breaking everything.)

opencensus/stats/internal/stats_manager_test.cc
isturdy: Done.

opencensus/stats/measure.h
isturdy: Added.

Issue #110, opened 2018-03-16
Intern tag keys (suggestion; breaks API)


isturdy:
Right now, tag keys are strings, and use string semantics. This has performance implications--we need to copy/hash/check equality frequently during aggregation.

We recommend that all keys be static constants to control visibility and avoid typos. Registering tag keys in the same way as measures (e.g. `static const opencensus::stats::TagKey method = opencensus::stats::RegisterTagKey("method");`) would encourage these best practices and offer performance benefits throughout the library.

PR #124, opened 2018-03-23
Add an interned TagKey class.

isturdy:
Implements #110.

### DIFF
--- a/opencensus/stats/internal/tag_key.cc
+++ b/opencensus/stats/internal/tag_key.cc
@@ -1,0 +1,73 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "opencensus/stats/tag_key.h"
+
+#include <cstdint>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "absl/strings/string_view.h"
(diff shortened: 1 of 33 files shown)


opencensus/stats/internal/tag_key.cc
```diff
@@ -0,0 +1,76 @@
...
+class TagKeyRegistry final {
+ public:
+  static TagKeyRegistry* Get() {
+    static TagKeyRegistry* global_tag_key_registry = new TagKeyRegistry();
+    return global_tag_key_registry;
+  }
+
+  TagKey Register(absl::string_view name);
+
+  // Return string_view rather than a const ref so that returned values will not
```
g-easy: But it returns a constref?

opencensus/stats/internal/tag_key_test.cc
```diff
@@ -0,0 +1,45 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "opencensus/stats/tag_key.h"
+
+#include "gmock/gmock.h"
```
g-easy: Don't need mock.

g-easy:
Please update commit description with text from issue #110 to explain why this is being done.

Is it worth including benchmark results here?

opencensus/stats/internal/tag_key.cc
isturdy: Removed comment. I went back on that because most of the uses required a std::string and the string_view interface thus caused a lot of copies. This should not be a problem in normal use where tag keys are initialized during startup.

isturdy:
Benchmark results:
```
Benchmark                                       Time             CPU      Time Old      Time New       CPU Old       CPU New
----------------------------------------------------------------------------------------------------------------------------
BM_RecordBatched_mean                        -0.0639         -0.0615          3646          3413          3634          3410
BM_DeltaProducerRecordBatched_mean           -0.1792         -0.1770           387           318           386           318

opencensus/stats/internal/tag_key_test.cc
isturdy: Done.

PR #125, opened 2018-03-23
Add a benchmark for RandomBuffer

isturdy:
Also update benchmark loop syntax to its present recommendation.


opencensus/common/internal/random_benchmark.cc
```diff
@@ -12,24 +12,36 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cstdint>
+
```
Vizerai: #include <vector>

isturdy: Done.

PR #127, opened 2018-03-23
Increase Stackdriver CreateTimeSeries batch size.

isturdy:
Stackdriver has fixed the backend limitations that constrained this previously. I ran the expanded e2e test locally.

### DIFF
--- a/opencensus/exporters/stats/stackdriver/internal/stackdriver_e2e_test.cc
+++ b/opencensus/exporters/stats/stackdriver/internal/stackdriver_e2e_test.cc
@@ -224,8 +224,9 @@ TEST_F(StackdriverE2eTest, LargeTest) {
       count_matchers;
 
   // The number of values to record for each tag--total metric cardinality will
-  // be tag_values^2.
-  const int tag_values = 10;
+  // be tag_values^2. We want this high enough that uploads require multiple
+  // batches.
+  const int tag_values = 25;
   for (int i = 0; i < tag_values; ++i) {
     for (int j = 0; j < tag_values; ++j) {
       const std::string tag1 = absl::StrCat("v1", i);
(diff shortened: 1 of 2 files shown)


opencensus/exporters/stats/stackdriver/internal/stackdriver_e2e_test.cc
```diff
@@ -225,7 +225,7 @@ TEST_F(StackdriverE2eTest, LargeTest) {
 
   // The number of values to record for each tag--total metric cardinality will
   // be tag_values^2.
-  const int tag_values = 10;
+  const int tag_values = 25;
```
Vizerai: Why 25?

isturdy: So that the number of time series pushed exceeds the batch size. 15+ would work, but this is fast enough.

g-easy: IMO worth adding this as a comment. :)

opencensus/exporters/stats/stackdriver/internal/stackdriver_exporter.cc
```diff
@@ -31,7 +31,8 @@ namespace {
 
 constexpr char kGoogleStackdriverStatsAddress[] = "monitoring.googleapis.com";
 constexpr char kProjectIdPrefix[] = "projects/";
-constexpr int kTimeSeriesBatchSize = 3;
+// Stackdriver limits a single CreateTimeSeries request to 200 series.
```
g-easy: If possible, please add URL which states this.

isturdy: I am not aware of one--the only place I have seen 200 is the error response when you try to upload more.

opencensus/exporters/stats/stackdriver/internal/stackdriver_e2e_test.cc
isturdy: Done.

PR #129, opened 2018-03-26
Update Travis and Appveyor's Bazel version.

isturdy:
I do not know of any issues with 0.8, but I think we may as well keep up.


g-easy:
Please also update the AppVeyor config, for consistency. :)

I don't know what gRPC's minimum bazel version requirement is, but if this comes up in future, we should probably have the same minimum.

PR #134, opened 2018-03-28
Change Stackdriver units to "1" for count views.

isturdy:
Counterpart to https://github.com/census-instrumentation/opencensus-java/pull/1087.


g-easy:
This seems reasonable. Do other languages do this? Can you think of any situation where we would want to count in units of something?

isturdy:
I know Java is, and I think Go is. As Bogdan noted in the Java PR, count is counting measurements, not the thing being measured--if someone wants to count in other units, they should use Sum with unit recorded values (as for the gRPC count measures/views).

PR #137, opened 2018-03-29
Bring back SpanExporter::ExportForTesting.

g-easy:
- Make StdoutExporter take an ostream.

- Make stdout_exporter_test faster (no sleep) and check that
  the added annotation appears in the output.


g-easy:
I'm not sure it's worth renaming, but I thought about it.

PR #138, opened 2018-03-30
Update the gRPC stats based on the new spec.

isturdy:
See https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md.


opencensus/plugins/grpc/internal/measures.cc
```diff
@@ -33,117 +33,87 @@ constexpr char kCount[] = "1";
...
   return measure;
 }
 
 stats::MeasureDouble RpcClientRoundtripLatency() {
-  static stats::MeasureDouble measure = stats::MeasureDouble::Register(
-      kRpcClientRoundtripLatencyMeasureName, "RPC roundtrip latency msec",
+  static const auto measure = stats::MeasureDouble::Register(
+      kRpcClientRoundtripLatencyMeasureName,
+      "Time between first byte of request sent to last byte of response "
+      "received, or terminal error.",
```
g-easy: This is the only description ending in a full stop. Suggest remove it for consistency.

opencensus/plugins/grpc/internal/server_filter.cc
```diff
@@ -187,16 +186,13 @@ void CensusServerCallData::Destroy(grpc_call_element *elem,
...
-        final_info->final_status == GRPC_STATUS_OK ? 0 : 1},
-       {RpcServerRequestBytes(), static_cast<double>(request_size)},
-       {RpcServerResponseBytes(), static_cast<double>(response_size)},
-       {RpcServerServerElapsedTime(), elapsed_time_ms},
-       {RpcServerRequestCount(), sent_message_count_},
-       {RpcServerFinishedCount(), 1},
-       {RpcServerResponseCount(), recv_message_count_}},
-      {{MethodTagKey(), qualified_method_},
-       {StatusTagKey(), StatusCodeToString(final_info->final_status)}});
+      {{RpcServerSentBytesPerRpc(), static_cast<double>(request_size)},
```
g-easy: From the server's point of view, isn't request size = recvd bytes?

opencensus/plugins/grpc/internal/measures.cc
isturdy: Done.

opencensus/plugins/grpc/internal/server_filter.cc
isturdy: Good point--switched the variable names.

PR #139, opened 2018-03-31
OpenCensus Zipkin Exporter

Vizerai:
Zipkin exporter for opencensus-cpp

- Uses HTTP protocol with libcurl
- Uses JSON encoding with rapidjson


Vizerai:
Recreating PR because CLA got stuck.

Vizerai:
I verified that it correctly exports the spans to the Zipkin server.  I added a README.md file with screenshots of the trace from the Zipkin UI.

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,386 @@
...
+constexpr char ipv4_loopback[] = "127.0.0.1";
+constexpr char ipv6_loopback[] = "::1";
+
+std::string SerializeMessageEvent(
+    const ::opencensus::trace::exporter::MessageEvent& event) {
+  return absl::StrCat(
+      event.type() == ::opencensus::trace::exporter::MessageEvent::Type::SENT
+          ? "SENT"
+          : "RECEIVED",
+      "/", std::to_string(event.id()), "/",
```
g-easy: What does event.id() return that can't be StrCatted without the explicit to_string()?

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
```diff
@@ -0,0 +1,386 @@
...
+    }
+    if ((res = curl_easy_setopt(curl, CURLOPT_MAXREDIRS,
+                                options.max_redirect_times)) != CURLE_OK) {
+      // Failed to set max redirect times.
+      return res;
+    }
+  }
+
+  // Sending HTTP request to url.
+  if ((res = curl_easy_perform(curl)) != CURLE_OK) {
```
g-easy: Can make this unconditional.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
```diff
@@ -0,0 +1,79 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef OPENCENSUS_EXPORTERS_TRACE_ZIPKIN_ZIPKIN_EXPORTER_H_
+#define OPENCENSUS_EXPORTERS_TRACE_ZIPKIN_ZIPKIN_EXPORTER_H_
+
+#include <memory>
+#include <string>
+
+#include "absl/memory/memory.h"
```
g-easy: I don't think memory or absl/memory are needed anymore. Ditto span_exporter and span

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: Fixed.

opencensus/exporters/trace/zipkin/zipkin_exporter.h
Vizerai: span.h -> removed.
absl/memory/memory.h -> needed for WrapUnique, moved to .cc file
span_exporter.h -> needed for handler and SpanData definitions, moved to .cc file

opencensus/exporters/trace/zipkin/internal/zipkin_exporter.cc
Vizerai: Done.

g-easy:
@adriancole Thanks for trying to build. It looks like you ran into https://github.com/grpc/grpc/issues/13856

We (opencensus-cpp) don't yet have CI on Mac OS X unfortunately.

The grpc issue is pretty old (January) so I'm surprised you ran into this. If you update grpc (with `bazel clean --expunge`) does that fix the problem?

adriancole:
thanks for the tip. got a bit closer..

INFO: Analysed 68 targets (59 packages loaded).

INFO: Found 68 targets...

*ERROR: */private/var/tmp/_bazel_acole/22ba4095a819d81558ab6179f365635a/external/com_github_curl/BUILD.bazel:11:1:
C++ compilation of rule '@com_github_curl//:curl' failed (Exit 1)

*external/com_github_curl/lib/if2ip.c:47:12: **fatal error: **'stropts.h'
file not found*

#  include <stropts.h>

*           ^~~~~~~~~~~*

1 error generated.

INFO: Elapsed time: 104.055s, Critical Path: 3.95s

*FAILED:* Build did NOT complete successfully

adriancole:
I'll try things here again
https://github.com/grpc/grpc/blob/master/INSTALL.md#macos just in case

g-easy:
The curl error is due to opencensus rather than grpc. We need to do some work on Mac OS X porting, by the looks of it.

PR #145, opened 2018-04-12
We no longer have to filter out targets with underscores.

g-easy:
Since the cc_grpc_library() rule was fixed in
https://github.com/grpc/grpc/pull/14625

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,5 @@ before_install:
 script:
   # We can't use --noshow_progress because Travis terminates the
   # build after 10 mins without output.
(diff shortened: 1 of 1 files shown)


g-easy:
This doesn't work yet. :(

ERROR: /home/travis/build/census-instrumentation/opencensus-cpp/google/rpc/BUILD:24:1: Couldn't build file google/rpc/_status_only-descriptor-set.proto.bin: Generating Descriptor Set proto_library //google/rpc:_status_only failed (Exit 1)

PR #147, opened 2018-04-16
Fix the curl build on os x.

isturdy:
`bazel test opencensus/exporters/trace/zipkin/...` passes on OS X.


adriancole:
thanks! will try later today

PR #158, opened 2018-05-15
Add test for stats stdout exporter.

g-easy:


opencensus/stats/BUILD
```diff
@@ -43,7 +43,7 @@ cc_library(
     srcs = ["testing/test_utils.cc"],
     hdrs = ["testing/test_utils.h"],
     copts = DEFAULT_COPTS,
-    visibility = ["//visibility:public"],
+    visibility = ["//opencensus:__subpackages__"],
```
isturdy: This was originally public so that 3rd-party exporters could write tests using synthetic ViewData (the only way of getting a ViewData with the rest of the public API is creating a view, recording, and then retrieving data).

g-easy: Reverted. Thanks.

PR #167, opened 2018-07-03
Benchmark has its own BUILD file as of v1.4.0.

g-easy:


g-easy:
I tried to manually confirm it works with: `$ bazel run //opencensus/common/internal:random_benchmark`

g-easy:
https://github.com/google/benchmark/issues/191

Issue #79, opened 2018-03-01
Fallthrough return in functions returning from a switch without a default.


isturdy:
Our build is presently cluttered with warnings that functions that return from a switch on an enum. Absl uses a return outside the enum to suppress these warnings; I think we should either follow suit (possibly with an `assert(false)`) or add build options to silence the warning.

g-easy:
These warnings come from gcc not being good enough at static analysis. I don't want to add too much noise to our source (or the generated code) just to quiet these warnings.

g-easy:
I agree on not silencing genuine errors.

How does `assert(false);` affect the generated code?

If the answer is "non zero", how about `__builtin_unreachable`?

isturdy:
`assert(false)` does still add a call in GCC's -O2 assembly. `__builtin_unreachable` does not, but I think making that portable would involve even worse clutter.

Seems like there is no great solution for this--I am okay with dropping it, particularly with all the warnings the proto and gRPC codegens create.

g-easy:
IMO keep this issue around at low priority in case someone else has a better idea. Thanks for the gcc followup.

On a related note, I like to leave out the `default:` case because it turns a missing case into a compile-time error. `default` potentially silences a genuine error.

isaachier:
I've wondered about this and decided to use this "pattern" in general:

```
enum { a, b, c };

switch (x) {
case a:
... 
case b:
... 
default:
  assert(x == c);
} 
```

Also, GCC isn't the issue. C++ allows for casting any value to an enum so the cases could never include these values outside the normal enum range.

isturdy:
That causes false alarms for the warning "enum not handled in switch", which typically does need attention, and is very helpful for catching omissions when adding an enum value.

isaachier:
What about the case label immediately followed by default and assertion? I agree in general it's nice to see all values but not sure there's a great way to deal with this for GCC.

PR #169, opened 2018-07-10
assert(false) after switch()

g-easy:
This stops gcc from warning about -Wreturn-type.
Make -Wreturn-type and -Wswitch into errors.
Fixes https://github.com/census-instrumentation/opencensus-cpp/issues/79


g-easy:
Merging #170 picked up this change so we can close this PR.

PR #177, opened 2018-09-05
Add StackdriverExporterOptions to StackdriverExporter::Register().

g-easy:
Use this to configure RPC deadlines and whether to block until
the exporter is connected.

Also change Register() to return a grpc::Status.

### DIFF
--- a/opencensus/exporters/trace/stackdriver/internal/stackdriver_exporter.cc
+++ b/opencensus/exporters/trace/stackdriver/internal/stackdriver_exporter.cc
@@ -246,10 +246,12 @@ void ConvertSpans(
 class Handler : public ::opencensus::trace::exporter::SpanExporter::Handler {
  public:
   Handler(absl::string_view project_id,
-          const std::shared_ptr<grpc::Channel>& channel)
+          const std::shared_ptr<grpc::Channel>& channel,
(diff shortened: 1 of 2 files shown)


opencensus/exporters/trace/stackdriver/internal/stackdriver_exporter.cc
```diff
@@ -277,11 +280,23 @@ void Handler::Export(
 
 }  // namespace
 
-void StackdriverExporter::Register(absl::string_view project_id) {
+grpc::Status StackdriverExporter::Register(
+    absl::string_view project_id, const StackdriverExporterOptions& opts) {
```
bogdandrutu: project_id should be part of the options.

See for example java:
https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverTraceConfiguration.java

Or in Go:
https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/stackdriver.go#L73

bogdandrutu:
Is this PR still needed?

g-easy:
This work has stalled for now, but I intend to finish it. What's left to do is: wait for the exporter to connect and return a grpc::Status.

PR #179, opened 2018-09-14
Add TagSet::DebugString().

g-easy:


g-easy:
```
$ bazel run opencensus/stats:tag_set_test
[...]
{"key1": "value1", "key2": "value2"}
```

opencensus/stats/internal/tag_set.cc
```diff
@@ -61,5 +63,17 @@ bool TagSet::operator==(const TagSet& other) const {
   return tags_ == other.tags_;
 }
 
+std::string TagSet::DebugString() const {
```
bogdandrutu: Add a comment that the format is not guaranteed so people cannot depend on this to retrieve tags.

g-easy: That comment is already in the header. Should I repeat it here?

PR #181, opened 2018-09-19
Travis: cache bazel builds between runs.

g-easy:


g-easy:
This reduces Travis walltime (on full cache hit) from ~10 mins to ~1 min.

PR #191, opened 2018-09-20
Mark all Span methods const.

g-easy:
Also, improve documentation and be explicit about copies and moves.

### DIFF
--- a/opencensus/trace/span.h
+++ b/opencensus/trace/span.h
@@ -157,7 +157,7 @@ class Span final {
   bool IsRecording() const;
 
  private:
-  Span();
+  Span() = delete;
   Span(const SpanContext& context, SpanImpl* impl);
 
   // Returns span_impl_, only used for testing.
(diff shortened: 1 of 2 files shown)


opencensus/trace/span.h
```diff
@@ -142,7 +157,7 @@ class Span final {
   bool IsRecording() const;
 
  private:
-  Span() {}
+  Span();
```
bogdandrutu: Why do we have this? Please comment if this is needed why.

opencensus/trace/span.h
```diff
@@ -153,7 +168,8 @@ class Span final {
   const SpanContext context_;
 
   // Shared pointer to the underlying Span representation. This is nullptr for
-  // Spans which are not recording events.
+  // Spans which are not recording events. This is an implementation detail, not
+  // part of the public API.
   std::shared_ptr<SpanImpl> span_impl_;
```
bogdandrutu: Why is this not const? Based on the current implementation this can be const. Until we need it to not be please make it const.

opencensus/trace/span.h
g-easy: Deleted.

opencensus/trace/span.h
g-easy: Mode const.

PR #193, opened 2018-09-20
Add operator!= to SpanContext.

g-easy:
Motivated by EXPECT_NE() in unit tests.


opencensus/trace/internal/span_context.cc
```diff
@@ -23,6 +23,10 @@ bool SpanContext::operator==(const SpanContext& that) const {
   return trace_id_ == that.trace_id() && span_id_ == that.span_id();
```
bogdandrutu: should also be the same options?

g-easy: We decided to exclude options, the equality is only checking whether the two contexts refer to the same span.

PR #194, opened 2018-09-21
Change our code to use the new tags:: namespace.

g-easy:


g-easy:
I tested this with `env CC=clang-5.0 tools/presubmit.sh` and searching the output for deprecation notices. There aren't any in our code.

PR #200, opened 2018-09-26
Draft implementation of Context.

g-easy:
Context holds information specific to an operation, such as a TagMap and
Span. Each thread has a currently active Context. Contexts are conceptually
immutable: the contents of a Context cannot be modified in-place.

### DIFF
--- a/opencensus/context/context.h
+++ b/opencensus/context/context.h
@@ -1,0 +1,69 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef OPENCENSUS_CONTEXT_CONTEXT_H_
+#define OPENCENSUS_CONTEXT_CONTEXT_H_
+
+#include <functional>
+#include <string>
+
+#include "opencensus/tags/tag_map.h"
+#include "opencensus/trace/span.h"
+
+namespace opencensus {
+namespace context {
+
+// Context holds information specific to an operation, such as a TagMap and
+// Span. Each thread has a currently active Context. Contexts are conceptually
+// immutable: the contents of a Context cannot be modified in-place.
+//
+// This is a draft implementation of Context, and we chose to depend on TagMap
+// and Span directly. In future, the implementation will change, so only rely
+// on the public API for manipulating Contexts. In future we may support
+// arbitrary keys and values.
+class Context {
+ public:
+  // Creates a default Context.
+  Context();
+
+  // Returns a const reference to the current (thread local) Context.
+  static const Context& Current();
+
+  // Context is copiable and movable.
+  Context(const Context&) = default;
+  Context(Context&&) = default;
+  Context& operator=(const Context&) = default;
+  Context& operator=(Context&&) = default;
+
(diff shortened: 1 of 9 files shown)


opencensus/context/internal/with_context_test.cc
```diff
@@ -0,0 +1,66 @@
...
+  opencensus::context::Context ctx;
+  opencensus::context::WithContext wc(ctx);
+}
+
+TEST(WithContextTest, WithContextMovable) {
+  opencensus::context::Context ctx;
+  opencensus::context::WithContext wc(std::move(ctx));
+}
+
+#if 0
```
bogdandrutu: Comment why 0?

opencensus/context/context.h
```diff
@@ -0,0 +1,69 @@
...
+// immutable: the contents of a Context cannot be modified in-place.
+//
+// This is a draft implementation of Context, and we chose to depend on TagMap
+// and Span directly. In future, the implementation will change, so only rely
+// on the public API for manipulating Contexts. In future we may support
+// arbitrary keys and values.
+class Context {
+ public:
+  // Creates a default Context.
+  Context();
```
bogdandrutu: Need this public? Please make it private until we need it.

opencensus/context/with_context.h
```diff
@@ -0,0 +1,52 @@
...
+#define OPENCENSUS_CONTEXT_WITH_CONTEXT_H_
+
+#include "opencensus/context/context.h"
+
+namespace opencensus {
+namespace context {
+
+// WithContext is an RAII object, only ever stack-allocate it. While it's in
+// scope, the execution will happen under a copy of the given Context.
+class WithContext {
```
bogdandrutu: We probably need a version of this that accepts a boolean condition to avoid duplicate code:

if (condition) {
  WithContext wc(ctx);
  ... // code
} else {
  ... // same code
}

opencensus/context/with_context.h
```diff
@@ -0,0 +1,52 @@
...
+  ~WithContext();
+
+ private:
+  WithContext() = delete;
+  WithContext(const WithContext&) = delete;
+  WithContext(WithContext&&) = delete;
+  WithContext& operator=(const WithContext&) = delete;
+  WithContext& operator=(WithContext&&) = delete;
+
+  Context swapped_context_;
```
bogdandrutu: Based on some experience that I have from dealing with other context implementation it is better if we have a way to check that objects are allocated and deallocated on the same thread. I suggest to put this check under the NDEBUG macro.

opencensus/context/internal/with_context_test.cc
g-easy: Done.

opencensus/context/context.h
g-easy: Done.

opencensus/context/with_context.h
g-easy: Filed https://github.com/census-instrumentation/opencensus-cpp/issues/202

opencensus/context/with_context.h
g-easy: Filed https://github.com/census-instrumentation/opencensus-cpp/issues/203

opencensus/trace/span.h
```diff
@@ -163,15 +166,19 @@ class Span final {
...
   // Spans that aren't sampled still have a valid SpanContext that propagates,
   // but no span_impl_.
-  const SpanContext context_;
+  SpanContext context_;
 
   // Shared pointer to the underlying Span representation. This is nullptr for
   // Spans which are not recording events. This is an implementation detail, not
   // part of the public API.
-  const std::shared_ptr<SpanImpl> span_impl_;
+  std::shared_ptr<SpanImpl> span_impl_;
```
g-easy: We made the shared_ptr non-const (and also the SpanContext above) so that we can use `swap()` (L170) and `operator=` (#191) to modify a Span object in-place, so that Context can have a Span member (as opposed to pointer to Span).

Also in (#191) we marked all Span methods const because:
1. They don't act on the shared_ptr, they act on the pointee.
2. In future we can replace the shared_ptr with something like opencensus-java's approach.

PR #201, opened 2018-09-26
examples/grpc: use gethostname() to determine hostname.

g-easy:


g-easy:
Even works on OS X now. :^)

PR #206, opened 2018-09-27
Check that WithContext is constructed and destructed on the same thread.

### LINKED ISSUE #203: Add thread debugging to WithContext.
Followup on https://github.com/census-instrumentation/opencensus-cpp/pull/200: add check under !NDEBUG to WithContext to make sure it's allocated and deallocated in the same thread.

g-easy:
This only happens in debug mode (!NDEBUG).
Add a death test for it.


g-easy:
Removed "fixed"

PR #207, opened 2018-09-27
Make WithContext conditional.

### LINKED ISSUE #202: Make WithContext() conditional.
Followup for https://github.com/census-instrumentation/opencensus-cpp/pull/200

g-easy:
While here, add an example in the comments.

### DIFF
--- a/opencensus/context/with_context.h
+++ b/opencensus/context/with_context.h
@@ -21,15 +21,22 @@ namespace opencensus {
 namespace context {
 
 // WithContext is a scoped object that sets the current Context to the given
-// one, until the WithContext object is destroyed.
+// one, until the WithContext object is destroyed. If the condition is false, it
+// doesn't do anything.
 //
-// Because it changes the current (thread local) context, NEVER allocate a
-// WithContext in one thread and deallocate in another. A simple way to ensure
-// this is to only ever stack-allocate it.
+// Because WithContext changes the current (thread local) context, NEVER
+// allocate a WithContext in one thread and deallocate in another. A simple way
+// to ensure this is to only ever stack-allocate it.
+//
+// Example usage:
+// {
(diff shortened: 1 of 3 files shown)


opencensus/context/with_context.h
```diff
@@ -21,15 +21,22 @@ namespace opencensus {
...
+// {
+//   WithContext wc(op.ctx_);
+//   // Do work.
+// }
 class WithContext {
  public:
-  explicit WithContext(const Context& ctx);
-  explicit WithContext(Context&& ctx);
+  explicit WithContext(const Context& ctx, bool cond = true);
+  explicit WithContext(Context&& ctx, bool cond = true);
```
bogdandrutu: Does this need to be public?

opencensus/context/with_context.h
```diff
@@ -39,6 +46,9 @@ class WithContext {
   WithContext& operator=(const WithContext&) = delete;
   WithContext& operator=(WithContext&&) = delete;
 
+  void ConditionalSwap();
+
+  const bool cond_;
```
bogdandrutu: Order the members by size.

opencensus/context/with_context.h
g-easy: Do you mean WithContext in general? If we want users to be able to capture contexts and install them later without using Wrap() then yes.

opencensus/context/with_context.h
g-easy: Done.

opencensus/context/with_context.h
g-easy: Bogdan meant the rvalue reference constructor.

The reason for keeping this is: if the caller has saved a copy of a Context, and wants to install it, and won't need it later, they can std::move() it to save another copy.

PR #210, opened 2018-09-28
WithSpan cleanups.

g-easy:


bogdandrutu:
I think this is more than cleanup. Please update description or the PR

g-easy:
PTAL. Much simpler diff after the merges.

PR #217, opened 2018-10-11
Add ScopedTags.

g-easy:


g-easy:
This is just a draft of the API, the implementation isn't there yet. Please let me know what you think.

opencensus/tags/scoped_tags.h
```diff
@@ -0,0 +1,58 @@
...
+#include <utility>
+
+#include "absl/strings/string_view.h"
+#include "opencensus/tags/tag_key.h"
+#include "opencensus/tags/tag_map.h"
+
+namespace opencensus {
+namespace tags {
+
+// ScopedTags is an RAII object, only ever stack-allocate it. While it's in
```
helly25: is a RAII... While it is in

opencensus/tags/scoped_tags.h
```diff
@@ -0,0 +1,58 @@
...
+// scope, the current Context will execute with the specified tags added. If a
+// tag key was already in use, its value will be replaced.
+class ScopedTags {
+ public:
+  explicit ScopedTags(
```
bogdandrutu: Looking at other languages:
https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/tags/TagContextBuilder.java
https://github.com/census-instrumentation/opencensus-go/blob/master/tag/map.go

It seems to me that we need some possible mutations:
* upsert/put which adds if not exist + update if exists.
* remove which removes a tag.

Should we support these operations? Should we also offer them to modify a TagMap (not in the context)?

g-easy: ScopedTags does upsert.

We don't currently have a way to delete, and TagMap's API is pretty limited (I'm working on this)

g-easy:
Opened issue #223 to track this.

PR #225, opened 2018-10-12
Add a variant of Record() that uses the currently active TagMap.

g-easy:


g-easy:
The TLS fix in #228 should make AppVeyor pass again on this PR.

PR #229, opened 2018-10-16
curl.bzl: Add Windows defines.

g-easy:


g-easy:
This is based on the output of running `nmake /f Makefile.vc mode=static` on libcurl. It gets us closer to having all exporters build on Windows.

PR #234, opened 2018-11-05
Stackdriver exporters: set user agent on gRPC channel.

g-easy:


WORKSPACE
```diff
@@ -37,6 +37,62 @@ http_archive(
     urls = ["https://github.com/google/benchmark/archive/master.zip"],
 )
 
+# gRPC and Prometheus supply different BUILD files for zlib.
```
g-easy: Ignore this, it's the build fix from #233

g-easy:
Please note that the generated user-agent string (I verified this with the gRPC client+server example) looks like:

```
opencensus-cpp [0.3.0-dev] grpc-c++/1.17.0-dev grpc-c/7.0.0-dev (linux; chttp2; gizmo)
```

Specifically: `opencensus-cpp [0.3.0-dev]` versus `opencensus-cpp/0.3.0-dev` with a slash.

g-easy:
Thanks for the approval! Does the user-agent string look correct though?

rakyll:
Looks correct to me, cc @songy23.

rakyll:
Not sure if space between opencensus-cpp and version is a correct format. User agent is a multi part header where separator is a space. @songy23, any reasons why we left a space between language and version?

rakyll:
We can simply can go with "opencensus-cpp/0.3.0-dev", but need to update the user agent in all other language implementations too.

g-easy:
I think the question was intended to be: how did we end up with the "name [ver]" format in the first place?

PR #235, opened 2018-11-07
examples/grpc: send and display gRPC metadata.

g-easy:


examples/grpc/hello_server.cc
```diff
@@ -60,6 +61,10 @@ opencensus::tags::TagKey CaseKey() {
   return key;
 }
 
+absl::string_view sv(const ::grpc::string_ref &s) {
```
helly25: Please do not make this "sv" but rather "to_string_view".

PR #236, opened 2018-11-13
Fix the build: remap prometheus's import of zlib to match gRPC's.

g-easy:


g-easy:
Waiting for a reply on https://github.com/jupp0r/prometheus-cpp/pull/225 to see if #236 should be merged or closed.

g-easy:
I don't know how long it will take to hash out jupp0r/prometheus-cpp#225 so I'm going to merge this now to get the build fixed.

Issue #175, opened 2018-09-02
No way to tell if StackdriverExporter::Register was successful


arfaian:
Ideally, the Register method for the StackdriverExporter would return a Status object.  Looking at the code, this may not be trivial to implement since the registration is actually handled in a separate thread.  Even so, it would still be useful to allow client code to block until registration is complete (either successful or unsuccessful) and take some action based on the result.

PR #240, opened 2018-11-23
Change the Stackdriver exporters' Register() to return Status.

g-easy:
Make this non-breaking change to the API now, so we can use it
to implement https://github.com/census-instrumentation/opencensus-cpp/issues/175
later.


g-easy:
I think these functions should return `grpc::Status` because almost all of the possible errors come from the RPC stack.

The other kind of Status we have in OpenCensus is for describing the status of a span in our own data model, which I don't think is appropriate here.

Abseil may one day release its own Status type, but that hasn't happened yet.

PR #242, opened 2018-11-26
Unbreak caching on Travis.

g-easy:
bazel 0.17.1 ignores HOME/.bazelrc (where we configure caching) if WORKSPACE/.bazelrc is present (which I populated in https://github.com/census-instrumentation/opencensus-cpp/commit/877550f3d72ef31aad662b1baf9d070fc22ce1b9 and broke caching).

Upgrade to bazel 0.19.2 which doesn't have this problem.


g-easy:
Build with:
empty cache: 15 mins
100% hit rate: 1 min

https://travis-ci.org/g-easy/opencensus-cpp/jobs/459585909
https://travis-ci.org/g-easy/opencensus-cpp/jobs/459588588

PR #252, opened 2018-12-05
Fix tools/format.sh: we were skipping a lot of files.

g-easy:
* Ignore CMake's ".build" dir.
* Improve CMake formatting while we're here.
* Reformat affected files.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -64,7 +64,7 @@ bind(
 # Used by prometheus-cpp.
 local_repository(
     name = "net_zlib_zlib",
-    path = "tools/zlib"
+    path = "tools/zlib",
 )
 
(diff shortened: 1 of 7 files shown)


WORKSPACE
```diff
@@ -64,7 +64,7 @@ bind(
 # Used by prometheus-cpp.
 local_repository(
     name = "net_zlib_zlib",
-    path = "tools/zlib"
+    path = "tools/zlib",
```
helly25: this change isn't really necessary

g-easy: This is true in terms of Python syntax, but buildifier makes this change, and as a result it's necessary in order to pass CI. :)

PR #267, opened 2019-01-08
Use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR for includes

nvartolomei:
Allows including the project using `add_subdirectory` command.

`CMAKE_SOURCE_DIR` always resolves to the top level project, breaking includes.


g-easy:
I verified that this still works with https://github.com/census-ecosystem/opencensus-cpp-example

g-easy:
Thanks!

PR #279, opened 2019-02-12
Add TraceOptions::WithSampling() helper.

terencekwt:
Allows user to set the IsSampled bit without exposing the implementation of how the bits are stored inside TraceOptions.

### DIFF
--- a/opencensus/trace/trace_options.h
+++ b/opencensus/trace/trace_options.h
@@ -50,8 +50,8 @@ class TraceOptions final {
   // Copies the TraceOptions data to a buffer, which must hold kSize bytes.
   void CopyTo(uint8_t* buf) const;
 
-  // Creates a new copy of a TraceOptions with sampled bit set by is_sampled
-  static TraceOptions SetSampled(const TraceOptions& parent, bool is_sampled);
+  // Returns a new copy of this TraceOptions with sampled bit set by is_sampled
+  TraceOptions WithSampling(bool is_sampled) const;
 
  private:
   friend class SpanGenerator;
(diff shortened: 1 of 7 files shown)


opencensus/trace/trace_options.h
```diff
@@ -42,6 +42,10 @@ class TraceOptions final {
   // the process, e.g. to Stackdriver/Zipkin.
   bool IsSampled() const;
 
+  // Set IsSampled for the TraceOptions. Normally, this is invoked
+  // implicitly by StartSpan
+  void SetSampled(bool is_sampled);
```
bogdandrutu: Please don't do this. We want the option to be immutable. If you want a helper function add a static function:

TraceOption SetSampled(TraceOption parent)

terencekwt: Are you proposing a static function that takes just the TraceOption argument like
`TraceOption SetSampled(TraceOption parent)`
or an additional argument like
`TraceOption SetSampled(TraceOption parent, bool isSampled)`

Both of which will make a new copy, but the latter option allows an additional input. Thanks

bogdandrutu: Fine with the second option. Just want to make sure that the object is immutable. This way we don't need to do any synchronization, also once a Span is created with a TraceOption then it will be the same for the entire life.

terencekwt: Ack.

That makes sense. Thanks, I will implement the second option then.

terencekwt: updated. 

I think the CI failure is not related to my change though?
https://travis-ci.org/census-instrumentation/opencensus-cpp/builds/495589423?utm_source=github_status&utm_medium=notification

g-easy: Currently:
```
TraceOptions b = TraceOptions::SetSampled(a, false);
```

Can we change the name? Also, it doesn't need to be static. How about:
```
TraceOptions b = a.WithSampling(true);
```

g-easy: (CI failure was fixed in #282)

PR #280, opened 2019-02-18
Use C++11 thread_local instead of compiler extensions.

g-easy:


g-easy:
The `cmake` failure on Travis is unrelated.

bogdandrutu:
To pass the required travis build if the cmake is not stable please mark it as allowed failure see https://github.com/census-instrumentation/opencensus-java/blob/master/.travis.yml#L38

g-easy:
Thanks for the pointer. For now I've fixed the build but I'll keep this in mind in future.

PR #283, opened 2019-02-19
Add NOMINMAX definitions when compiling with MSVC

meastp:
This adds NOMINMAX as a public/interface definition to the targets `opencensus_common_stats_object` and `opencensus_stats_core`

### DIFF
--- a/opencensus/common/internal/CMakeLists.txt
+++ b/opencensus/common/internal/CMakeLists.txt
@@ -24,7 +24,7 @@ opencensus_lib(common_random
 
 opencensus_lib(common_stats_object DEPS absl::time)
 
-# Defines NOMINMAX when compiling with MSVC to fix build errors
+# Define NOMINMAX to fix build errors when compiling with MSVC.
 target_compile_definitions(opencensus_common_stats_object INTERFACE $<$<CXX_COMPILER_ID:MSVC>:NOMINMAX>)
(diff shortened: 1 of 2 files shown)


opencensus/common/internal/CMakeLists.txt
```diff
@@ -24,6 +24,8 @@ opencensus_lib(common_random
 
 opencensus_lib(common_stats_object DEPS absl::time)
 
+target_compile_definitions(opencensus_common_stats_object INTERFACE $<$<CXX_COMPILER_ID:MSVC>:NOMINMAX>)
```
g-easy: Should NOMINMAX apply to all of opencensus instead of just the two libraries in this PR?

Also, please add a comment about what this line is doing and why. :)

meastp: Previously I submitted NOMINMAX applied to the entire library, and the response I got was that this should be done only per target where required. Since this target is an INTERFACE target, the targets using this target will also define NOMINMAX.

g-easy: Ah, right, coryan@ asked for just the affected targets. Let's do that then.

How about:
```
# We need MSVC to not define min() and max() macros.
```

Also please run `tools/format.sh` to make CI pass.

PR #284, opened 2019-02-19
Add gtest_force_shared_crt for MSVC

meastp:


g-easy:
Please describe why this is being done.

meastp:
This is being done simply because the build (or, to be more specifik, the linking) on MSVC fails when abseil and opencensus-cpp compile with /MD, and googletest compiles with /MT.

>       # force this option to ON so that Google Test will use /MD instead of /MT
 >     # /MD is now the default for Visual Studio, so it should be our default, too

g-easy:
Makes sense. Thanks a lot for the followup. Please run `tools/format.sh`

meastp:
I'm on Windows, so I can not run tools/format.sh directly, but by reading the script I think only the following applies to my changes (I'm not familiar with the `buildify` command, so I don't know if it applies):

> # No CRLF line endings.
> sed -i 's/\r$//' $($FIND -type f -print)
> # No trailing spaces.
> sed -i 's/ \+$//' $($FIND -type f -print)

If this is correct I'll fix that. Please confirm that my understanding is correct... :)

g-easy:
Those definitely apply, but cmake-format might be in play for this PR as well.

g-easy:
It's not just whitespace. I did the format on my end but I don't think I can push it to this PR:

```
git checkout -b meastp-meastp_msvc_add_gtest_force_shared_crt master
git pull https://github.com/meastp/opencensus-cpp.git meastp_msvc_add_gtest_force_shared_crt
tools/format.sh
git commit -a -m format
git push g-easy meastp-meastp_msvc_add_gtest_force_shared_crt
```

Please apply this commit to the PR: https://github.com/g-easy/opencensus-cpp/commit/17cc8627f129f5a75b2f83a17af4be8e0e788f42

g-easy:
If you prefer, I can try to do it in the online editor. :)

meastp:
Thanks, I managed to do it online by creating a pull request with `compare` from your repository to my repository :)

PR #285, opened 2019-02-21
Get rid of TraceOptions::SetSampled() in favor of WithSampling().

g-easy:


WORKSPACE
```diff
@@ -42,8 +42,8 @@ http_archive(
 # gRPC
 http_archive(
     name = "com_github_grpc_grpc",
-    strip_prefix = "grpc-master",
-    urls = ["https://github.com/grpc/grpc/archive/master.tar.gz"],
+    strip_prefix = "grpc-b03e014ad82c39b7c91d794bc775c0d4ec187e1b",
```
bogdandrutu: Add a comment to use master in the future?

g-easy: Absolutely. But I want to do it in a separate commit: https://github.com/census-instrumentation/opencensus-cpp/pull/286

PR #290, opened 2019-02-28
CMake: Replace ExternalProject_Add with FetchContent

meastp:
This reduces the amount of code for including/building dependencies. ~~The only drawback is an increase in the CMake version required.~~


bogdandrutu:
@g-easy this looks good to me (from a person with 0 knowledge about cmake, but I like when code gets deleted). What do you think?

meastp:
We already discussed it in #244, but I'd like @g-easy to look at it anyway :)

g-easy:
Please make this pass on Travis. :)

meastp:
Well, I can't run the tools/format.sh script as I'm on Windows. Could you help me with that like last time? :)

The other failure is simply because the travis node's CMake version is too old:

> CMake Error at CMakeLists.txt:15 (cmake_minimum_required):
> CMake 3.11 or higher is required.  You are running version 3.9.2

bogdandrutu:
@meastp you can create a small docker image map your clone directory, and run the script :) maybe we should have a "tools/format_docker.sh" which actually is useful not only for your usecase but also to make sure that we run that specific clang-format.

meastp:
> 
> 
> @meastp you can create a small docker image map your clone directory, and run the script :) maybe we should have a "tools/format_docker.sh" which actually is useful not only for your usecase but also to make sure that we run that specific clang-format.

What if travis could just push a commit to the pull request with the required changes if the tools/format.sh fails? :)

https://gist.github.com/willprice/e07efd73fb7f13f917ea

g-easy:
Regular Xenial ships cmake 3.5.1. I got this running locally with ubuntu:cosmic. https://launchpad.net/ubuntu/+source/cmake

g-easy:
@coryan Is the cmake 3.11 requirement okay with you?

coryan:
And Ubuntu:bionic is 3.10, and let's not get started on CentOS.  I think 3.11 might be too high given the current versions on popular distros.

meastp:
What about including `FetchContent` directly, then, as was suggested in https://github.com/census-instrumentation/opencensus-cpp/pull/256#issuecomment-448406031?

g-easy:
I think `FetchContent.cmake` is worth a try. What is the minimum version requirement with this option?

meastp:
@coryan @g-easy It looks like this works with the old cmake versions. :) 

@g-easy Could you please fix format.sh for me?

Issue #289, opened 2019-02-27
MSVC version supported


meastp:
The current MSVC version specified is 19, but I don't think that is easy to deduce by everybody. (see https://devblogs.microsoft.com/cppblog/side-by-side-minor-version-msvc-toolsets-in-visual-studio-2017/)

MSVC 2017 has versions x..y.z where x is MSVC 2015, 2017 etc, y is the update number and z is the bugfix/patch-number.

I'm testing on MSVC 2017 15.9.7 (latest update and patch version)

g-easy:
Can you tell what Appveyor is using?

meastp:
According to the documentation: https://www.appveyor.com/docs/windows-images-software/#visual-studio-2017 -> 15.9.6.

15.9 is the most important bit - the last is just patch version and these are released quite often.

g-easy:
IMO we should specify this (and the travis clang and gcc versions) in the topmost README.md

"Support" so far has been a bit ad-hoc. We merge PRs that improve Windows support (e.g. #283) and also Sturdy and I spent some time trying to generally make things build, but not all of OpenCensus is built and tested on Appveyor yet. PRs welcome. :)

meastp:
Could you specify exactly what's missing for Windows / MSVC support? Perhaps a followup issue with a list of all known missing functionality, like with the cmake followup issue? :)

g-easy:
IMO what's missing is "build and test everything on Appveyor." As in `blaze test //...`

This issue is related to https://github.com/census-instrumentation/opencensus-cpp/issues/6

g-easy:
Is this issue resolved now that the MSVC version is specified? We still have #6 to track Windows CI.

PR #292, opened 2019-03-05
Fix Windows builds with appveyor

meastp:
This is WIP, ref #289 

I'm not familiar at all with bazel builds, so I'm starting with CMake. That's also an incentive to expand the CMake support ;)

It looks like appveyor has support for `vcpkg` as well, so it could be used to test a future vcpkg-build of opencensus with find_package for dependencies :)

That said, supporting both is an option: https://github.com/google/googletest/blob/master/appveyor.yml

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -22,8 +22,9 @@ test_script:
     
     Get-ChildItem -Recurse -Path $env:APPVEYOR_BUILD_FOLDER -Filter "*_test.exe" | ForEach-Object { 
       Write-Output $_.FullName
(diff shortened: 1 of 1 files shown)


meastp:
@g-easy @bogdandrutu As far as I can tell, this is complete. Is it really neccessary with Bazel build? Could that be done in another PR?

g-easy:
I thought it was gonna be easier with bazel. :^)

How about this:
* Get this PR in pretty much as-is.
* In a followup PR, could you please restore the bazel build?

PR #297, opened 2019-03-14
gRPC calls moved to grpc_impl namespace.

g-easy:
This fixes the build.

See also: https://github.com/grpc/grpc/pull/18223


g-easy:
@karthikravis does this seem ok? This is an example of what user code should look like.

PR #298, opened 2019-03-14
Don't autoformat FetchContent.cmake.

g-easy:
This file comes from upstream, so let's keep it as-is.
This file will be added in #290.


g-easy:
cc @meastp

PR #305, opened 2019-03-28
Add support for StrictSpans mode in Tracing

jhferris:
Goal: StrictSpans mode is meant to enforce the invariant that a parent span cannot
end while it has children running. This can happen if your service has an internally-asynchronous architecture.

In this solution, if a parent span ends with active children, it is held open
and its end is triggered when the last child ends. This works even with a chain of spans, the last child
can trigger all parents up to the root to end.

Details:
1. Switch running_span_store_impl to key its map on SpanId instead of the pointer
2. Add FindSpanById method on RunningSpanStoreImpl
3. Add reference count called active_child_count_ on SpanImpl to keep track of active children
4. Add logic at span start and end time to update the reference count on parents,
5. Add tests


jhferris:
I realize this is a nontrivial change. It started out as "I wonder how hard it would be to do this" and kind of grew from there. That said, I believe this solves an actual use case, and I've tried to make it so that the code changes/impacts are:

1) minimal if you're not running in StrictSpans mode (not the best name, I'm open to suggestions)
2) As clean/fitting into the existing structure of the code as possible
3) Tested :) 

I'm happy to split up the pull request into smaller pieces if that would be easier or make any other changes you feel are appropriate, if you think the core idea is viable.

g-easy:
Where does the StrictSpans idea come from? I can't find any references to this term.

g-easy:
This PR adds a new TraceOptions bit. Could you please open an issue at https://github.com/census-instrumentation/opencensus-specs and describe the motivating use case for "strict" mode. There are several implementations of OpenCensus beyond opencensus-cpp, and the project needs to decide whether to implement this feature in every language.

Also I'm worried about the performance overhead of the refcounting, and that we won't be able to replace the underlying implementation with a Disruptor-style approach (e.g. like what opencensus-java does).

PR #306, opened 2019-03-28
Update format.sh

jhferris:
TLDR probably shouldn't be formatting those to our style anyway, and there is a file with spaces in there that breaks stuff.

### DIFF
--- a/tools/format.sh
+++ b/tools/format.sh
@@ -20,7 +20,7 @@ if [[ ! -e tools/format.sh ]]; then
 fi
 set -e
 # Correct common miscapitalizations.
-FIND="find . -name .git -prune -o -name .build -prune -o"
+FIND="find . -name .git -prune -o -name _deps -prune -o -name .build -prune -o"
 sed -i 's/Open[c]ensus/OpenCensus/g' $($FIND -type f -print)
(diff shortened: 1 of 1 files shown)


g-easy:
Where does `_deps` come from?

The only mention I can find is in `cmake/FetchContent.cmake` which results in `.build/_deps` which should already be excluded on line 23 (where `.build` is excluded)

What am I missing?

jhferris:
Something in the build process makes a top level _deps directory:

ferrisj@feisbox:~/opencensus-cpp$ ag -g spaces
_deps/prometheus-src/3rdparty/civetweb/test/dir with spaces/hello.cgi

I got this from the error in CI from my other PR as well as trying to run it locally.

g-easy:
How are you running CMake? I was not able to repro this on my workstation with CMake 3.10.2 and following the instructions in `cmake/README.md`

jhferris:
I hadn't seen the README in the cmake dir (I'd checked the toplevel README and didn't see anything).

I was building with just `cmake .`

g-easy:
This seems like a common enough pattern that it's worth special-casing `_deps`.

PR #307, opened 2019-03-29
Fix absl cops flag names.

jinmel:
absl copts build flags has been updated to ABSL_* 

this fixes build error


g-easy:
Thanks for the build fix!

jinmel:
@g-easy many of the google product repositories use master branch of absl and absl make changes that causes build errors on other google products...

g-easy:
We hope this will improve in future. Also, as part of the release process, we pin opencensus-cpp dependencies to known-working versions.

PR #310, opened 2019-04-02
Added parameters to tools/format.sh (Windows)

meastp:
This adds parameters for skipping the sed operations and/or the operations that require installed software (clang-format, cmake-format ...).

The usage scenario is that @g-easy 's Docker-container is not useable in Windows Docker because the sed operations fail. However, the workaround is easy:
- Install Git for Windows - this gives you a bash shell. Run format.sh from here, skipping the operations requiring installed software
- Use @g-easy 's Docker container for the operations requiring installed software.

I have tested this, and it works very well :)

### DIFF
--- a/tools/format.sh
+++ b/tools/format.sh
@@ -18,37 +18,69 @@ if [[ ! -e tools/format.sh ]]; then
   echo "This tool must be run from the topmost OpenCensus directory." >&2
   exit 1
 fi
+
 set -e
-# Correct common miscapitalizations.
+
+usage()
+{
+    echo "usage: format.sh [--disable-sed-check] [--disable-tools-check]"
+}
+
+sed_check=y
+tools_check=y
+
+while [ "$1" != "" ]; do
+    case $1 in
+        --disable-sed-check )   sed_check=n
+                                ;;
+        --disable-tools-check ) tools_check=n
+                                ;;
+        -h | --help )           usage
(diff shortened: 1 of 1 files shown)


tools/format.sh
```diff
@@ -18,37 +18,69 @@ if [[ ! -e tools/format.sh ]]; then
...
+
+usage()
+{
+    echo "usage: format.sh [--disable-sed-check] [--disable-tools-check]"
+}
+
+sed_check=y
+tools_check=y
+
+while [ "$1" != "" ]; do
```
g-easy: `[[ ... ]]` for consistency with the rest of the file.

tools/format.sh
```diff
@@ -57,3 +89,9 @@ if [[ ! -z "$CHANGED" ]]; then
 else
   echo "No changes."
 fi
+
+if [ "$tools_check" == "$tools_check" ]; then
```
g-easy: This doesn't look right. Also it breaks Travis.

meastp: ups - should have been $sed_check == $tools_check :)

g-easy:
It passes on Travis now. @meastp are you happy with my edits / the state of this PR in general?

meastp:
Thank you for the fixes, @g-easy - LGTM :)

PR #314, opened 2019-04-15
Add HTTP example server.

g-easy:


g-easy:
@bogdandrutu What do you think?

PR #317, opened 2019-05-10
Pin cmake_format to version 0.4.5.

g-easy:
Version 0.5.1 changes the formatting (which breaks "check format" on
travis) but also breaks add_library(ALIAS) so we can't upgrade.

Upstream issue: https://github.com/cheshirekow/cmake_format/issues/111


g-easy:
cc @jinmel

PR #321, opened 2019-05-29
Add Span::DebugString() method

jinmel:


jinmel:
@g-easy I had to debug for some case where it failed to export. It happens frequently when creating span from remote propagation.

g-easy:
Was the SpanContext sufficient to debug the remote propagation problem? The TraceId and sampled bit are present at least.

jinmel:
I just had to make sure that grpc instrumentation is working properly.

opencensus/trace/internal/span.cc
```diff
@@ -221,7 +221,10 @@ bool Span::IsSampled() const { return context_.trace_options().IsSampled(); }
 bool Span::IsRecording() const { return span_impl_ != nullptr; }
 
 std::string Span::DebugString() const {
-  return span_impl_->ToSpanData().DebugString();
+  if (IsRecording()) {
+    return span_impl_->ToSpanData().DebugString();
+  }
+  return "Span has not started.";
```
g-easy: I think this should be "Span is not recording" because "start" has a specific meaning.

bogdandrutu:
@jinmel can you clarify a bit more? We would like to not commit to support this functionality in the long term, but I understand your need to debug this problem.

I have some questions:
1) Why a stdout exporter is not enough?
2) Do you need this functionality in production code or only when you try to debug locally during development?

jinmel:
@bogdandrutu 
1. Span was not being exported.
2. It is for development stage. 

I just think it is strange we have very limited view of span although we have a span class itself.

g-easy:
I agree, but we want to be able to replace the SpanImpl later, with something more efficient, and as a result we're very conservative about which APIs we're willing to add.

jinmel:
@g-easy I'll close this issue because it seems like this PR could introduce breaking changes in future based on your comments. I'd be glad to know if there is any future plans you are thinking of. 

Please leave a reference here if there is a change in Span class that might make this feature available to add. 

Thanks.

PR #325, opened 2019-06-04
Use proto best practices.

g-easy:
Add separate targets for:

 * proto_library
 * cc_proto_library
 * cc_grpc_library


g-easy:
cc @vam-google

PR #335, opened 2019-06-25
Add Span::SetName()

kyessenov:
Allows changing the span name after the constructor before the span end.

This is needed for envoyproxy/envoy which updates the span name during the request route configuration.
Currently, it uses an annotation which leads to strange span values:
```
Name: ingress (From Client)
TraceId-SpanId-Options: 09d76941cce57c762c600ad861e120e3-57f2c613dc9714a0-01
Parent SpanId: 91b716c92ee92e57 (remote: true)
Start time: 2019-06-24T13:47:07.150968383-07:00
End time: 2019-06-24T13:47:07.153027742-07:00
Attributes: (0 dropped)
Annotations: (0 dropped)
  2019-06-24T13:47:07.151008446-07:00: setOperation (attributes: "operation":"ServerRoute")
  2019-06-24T13:47:07.151011373-07:00: setOperation (attributes: "operation":"ClientRoute")
Message events: (0 dropped)
Links: (0 dropped)
Span ended: true
Status: OK
```

We should not be seeing the server operation in the client span. Instead, the span name should be ClientRoute. 

Signed-off-by: Kuat Yessenov <kuat@google.com>

### DIFF
--- a/opencensus/trace/internal/span.cc
+++ b/opencensus/trace/internal/span.cc
@@ -202,6 +202,12 @@ void Span::SetStatus(StatusCode canonical_code,
   }
 }
 
+void Span::SetName(absl::string_view message) const {
+  if (IsRecording()) {
+    span_impl_->SetName(message);
+  }
+}
+
 void Span::End() const {
   if (IsRecording()) {
     if (!span_impl_->End()) {
--- a/opencensus/trace/internal/span.cc
+++ b/opencensus/trace/internal/span.cc
@@ -202,9 +202,9 @@ void Span::SetStatus(StatusCode canonical_code,
   }
 }
 
-void Span::SetName(absl::string_view message) const {
+void Span::SetName(absl::string_view name) const {
   if (IsRecording()) {
-    span_impl_->SetName(message);
+    span_impl_->SetName(name);
   }
 }
 
--- a/opencensus/trace/internal/span_impl.h
+++ b/opencensus/trace/internal/span_impl.h
@@ -121,7 +121,7 @@ class SpanImpl final {
   // The status of the span. Only set if start_options_.record_events is true.
   exporter::Status status_ GUARDED_BY(mu_);
   // The displayed name of the span.
-  std::string name_;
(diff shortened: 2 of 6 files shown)


kyessenov:
Looks like grpc-master fails to build on CI, but it works locally for me. It might be related to bazel version. I have 0.24.1 but travis has 0.20.1. Can someone update travis definitions, or should I do that here?

opencensus/trace/internal/span.cc
```diff
@@ -202,6 +202,12 @@ void Span::SetStatus(StatusCode canonical_code,
   }
 }
 
+void Span::SetName(absl::string_view message) const {
```
bogdandrutu: Why called message?

opencensus/trace/internal/span_impl.h
```diff
@@ -119,7 +121,7 @@ class SpanImpl final {
   // The status of the span. Only set if start_options_.record_events is true.
   exporter::Status status_ GUARDED_BY(mu_);
   // The displayed name of the span.
-  const std::string name_;
+  std::string name_;
```
bogdandrutu: Need to be guarded_by

opencensus/trace/internal/span.cc
kyessenov: Oops, not sure what i was thinking. Corrected.

opencensus/trace/internal/span_impl.h
kyessenov: Done, thanks.

kyessenov:
@bogdandrutu 
I don't think I have a good explanation for why span name is not known a-priori. My understanding is that span name is set in the route, and route is not known when a request gets processed first (and that's when the tracing span is created). FWIW zipkin and opentracing allow settings the name on spans.

g-easy:
This is related to https://github.com/census-instrumentation/opencensus-specs/issues/217

g-easy:
Neat, thanks!

Issue #334, opened 2019-06-24
Stackdriver exporter: support exporting GCP metrics (MonitoredResource)


bianpengyuan:
Currently stackdriver exporter could only export custom metrics with `Global` MR. It seems easy to add support to export GCP metrics with predetermined MR, such as [Istio metrics](https://cloud.google.com/monitoring/api/metrics_anthos#anthos-istio.io). We can achieve this by extending stackdriver options to have monitored resource type and labels, which overwrite `Global` MR when constructing time series.

@g-easy

g-easy:
Should this configuration be per-view rather than in the exporter options?

https://github.com/census-instrumentation/opencensus-cpp/blob/a506cf846edca75b93e5457aca51c568378201be/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc#L184

c @bogdandrutu

g-easy:
Related issue: #291

bianpengyuan:
Per view would be better, though I don't know if it is appropriate to add stackdriver specific concept into view.

Also with ability to exporting GCP metrics, I think metric descriptor registration does not need to happen for every view right?

g-easy:
Alternatively, we could add per-view overrides to the exporter rather than the view.

Issue #291, opened 2019-03-01
Stackdriver export metric prefix


prateekr:
https://github.com/census-instrumentation/opencensus-java/blob/89ecd70f633be3e74cd9d9d20934804eebf49ca6/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsConfiguration.java#L138-L149

Similar to the Java version, it'd be quite useful for the stackdriver exporter in cpp to implement the metric name prefix as well.

PR #336, opened 2019-06-26
StackdriverOptions: add metric_name_prefix.

g-easy:


g-easy:
This addresses #334 and #291.

The change to `StackdriverOptions` is in line with that Java does, as pointed out by @prateekr: https://github.com/census-instrumentation/opencensus-java/blob/89ecd70f633be3e74cd9d9d20934804eebf49ca6/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsConfiguration.java#L129-L149

bianpengyuan:
Thanks for the PR! I am wondering does this fully resolve #334? Monitored resource labels are not being overwrite. Also what will happen when `CreateMetricDescriptor` is called for an existing GCP metric? Will it return failure and cause view registration to be skipped?

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
```diff
@@ -36,6 +36,12 @@ struct StackdriverOptions {
 
   // The RPC deadline to use when exporting to Stackdriver.
   absl::Duration rpc_deadline = absl::Seconds(5);
+
+  // If empty, the exporter will use "custom.googleapis.com/opencensus"
+  std::string metric_domain;
```
bogdandrutu: Please explain with probably an example of a metric how this will be used.

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
```diff
@@ -36,6 +36,12 @@ struct StackdriverOptions {
...
+
+  // If empty, the exporter will use "global"
+  std::string resource_type;
```
bogdandrutu: Do we have a logic to determine how the resource labels are set?

prateekr: Thoughts on using [MonitoredResource proto](https://github.com/googleapis/googleapis/blob/07244bb797ddd6e0c1c15b02b4467a9a5729299f/google/monitoring/v3/metric.proto#L60) directly in this config?  The [definition](https://github.com/googleapis/googleapis/blob/d84bde8698967faa93ab56cc7132965f6be534ad/google/api/monitored_resource.proto#L86) already has comments around its usage as well.

Otherwise we also need zone and instance_id here.

g-easy:
This doesn't fully resolve #334, it doesn't add labels, but it lets users start to override metadata.

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
g-easy: Added docs that make it look like the Java version, and also renamed to `metric_name_prefix` to match Java.

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
g-easy: I've split this change out of this PR. I'll do a followup PR with the proto as suggested.

g-easy:
@bogdandrutu PTAL.

PR #338, opened 2019-06-28
override sampling decision

kyessenov:
Clients of opencensus may want to delay the sampling decision until after the span has been created.
For example, at the time envoyproxy decides whether a request is a health check or not, it may already have started recording the span:
1. Envoy receives a request.
2. Envoy creates a child span for an external authorization sub-request.
3. Envoy applies a health check blacklist, may want to stop sampling the span and its children.
4. Envoy proceedes forwarding the request.

To support this use case, other tracers expose a `setSampling` override. I think it's sufficient to stop recording the span for this case, e.g. only override the sampling decision to false.

Signed-off-by: Kuat Yessenov <kuat@google.com>

### DIFF
--- a/opencensus/trace/internal/span.cc
+++ b/opencensus/trace/internal/span.cc
@@ -224,6 +224,8 @@ const SpanContext& Span::context() const { return context_; }
 
 bool Span::IsSampled() const { return context_.trace_options().IsSampled(); }
 
+void Span::StopRecording() { span_impl_ = nullptr; }
+
 bool Span::IsRecording() const { return span_impl_ != nullptr; }
 
 void swap(Span& a, Span& b) {
--- a/opencensus/trace/span.h
+++ b/opencensus/trace/span.h
@@ -168,6 +168,11 @@ class Span final {
   // Sampled spans always record events.
   bool IsSampled() const;
 
+  // Stops recording the span and overrides the sampling decision to false.
+  // This should be used by the client to force disable sampling in support of
+  // delayed sampling decisions (e.g. health check URI blacklist).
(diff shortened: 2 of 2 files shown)


kyessenov:
cc @objectiser

g-easy:
BTW we discussed this in https://github.com/census-instrumentation/opencensus-specs/issues/218

opencensus/trace/internal/span.cc
```diff
@@ -224,6 +224,8 @@ const SpanContext& Span::context() const { return context_; }
 
 bool Span::IsSampled() const { return context_.trace_options().IsSampled(); }
 
+void Span::StopRecording() { span_impl_ = nullptr; }
```
g-easy: This approach will not work, it prevents Span::End from doing:
```
exporter::RunningSpanStoreImpl::Get()->RemoveSpan(span_impl_);
```
(i.e. leaks memory)

g-easy:
How about this approach: https://github.com/census-instrumentation/opencensus-cpp/pull/339

@bogdandrutu what do you think?

opencensus/trace/internal/span.cc
kyessenov: Ouch. I was not aware of the mechanics of memory management here

kyessenov:
I'm fine with either approach. The health checking trace noise is the most visible impact of _not_ having span abandonment, which should be possible to fix with #339.

opencensus/trace/internal/span.cc
g-easy: I'm sorry. I wish things were simple.

PR #341, opened 2019-07-04
Stackdriver Trace exporter: allow override of trace_service_stub.

g-easy:


opencensus/exporters/trace/stackdriver/stackdriver_exporter.h
```diff
@@ -31,12 +33,18 @@ struct StackdriverOptions {
 
   // The RPC deadline to use when exporting to Stackdriver.
   absl::Duration rpc_deadline = absl::Seconds(5);
+
+  // (optional) By default, the exporter connects to Stackdriver using gRPC. If
+  // this stub is non-null, the exporter will use this stub to send gRPC calls
+  // instead. Useful for testing.
```
mandarjog: This is great!
In the envoy OC tracer I would like to see envoys [google](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/grpc_service.proto#envoy-api-msg-core-grpcservice-googlegrpc) grpc stub and service endpoint as additional options.

opencensus/exporters/trace/stackdriver/stackdriver_exporter.h
```diff
@@ -31,12 +33,18 @@ struct StackdriverOptions {
...
+  std::unique_ptr<google::devtools::cloudtrace::v2::TraceService::Stub>
+      trace_service_stub;
 };
 
 class StackdriverExporter {
  public:
   // Registers the exporter.
-  static void Register(const StackdriverOptions& opts);
+  static void Register(StackdriverOptions&& opts);
```
bogdandrutu: Is this not a breaking change? What about just accept pass by value?

g-easy: Yeah it's a breaking change - see the diffs in dir examples/grpc/...

We can't pass StackdriverOptions by value if there's a unique_ptr in there.

We could pass it by ref and swap out the contents. (option 1)

Or put a plain pointer in StackdriverOptions and document it as transferring ownership. (option 2)

What do you think?

g-easy:
I updated the PR. This is what it would look like to pass by ref. What do you think?

opencensus/exporters/trace/stackdriver/internal/stackdriver_exporter.cc
```diff
@@ -261,29 +256,52 @@ void Handler::Export(
...
+// static
+void StackdriverExporter::Register(StackdriverOptions&& opts) {
+  if (opts.trace_service_stub == nullptr) {
+    opts.trace_service_stub = MakeStackdriverStub();
+  }
+  ::opencensus::trace::exporter::SpanExporter::RegisterHandler(
+      absl::make_unique<Handler>(std::move(opts)));
+}
+
+// static, DEPRECATED
```
bogdandrutu: Does absl provide a deprecated macro?

g-easy: Yep, I use it in the header: https://github.com/census-instrumentation/opencensus-cpp/pull/341/files#diff-a00bc854d48bc0ac9e92a7c77d2f4d0eR51

PR #344, opened 2019-07-14
Add per view monitored resource into stackdriver exporter option

bianpengyuan:
Follow up of #334. In addition to general custom monitored resource, also add per view custom monitored resource. This could be useful in the case of using envoy exporting istio metric, where inbound metrics attached to k8s_container MR and outbound metrics attached to k8s_pod (https://cloud.google.com/monitoring/api/metrics_anthos).

### DIFF
--- a/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
+++ b/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
@@ -190,10 +190,9 @@ std::vector<google::monitoring::v3::TimeSeries> MakeTimeSeries(
   auto base_time_series = google::monitoring::v3::TimeSeries();
   base_time_series.mutable_metric()->set_type(
       MakeType(metric_name_prefix, view_descriptor.name()));
-  if (per_view_monitored_resource.find(view_descriptor.name()) !=
-      per_view_monitored_resource.end()) {
-    *base_time_series.mutable_resource() =
-        per_view_monitored_resource.at(view_descriptor.name());
+  auto iter = per_view_monitored_resource.find(view_descriptor.name());
+  if (iter != per_view_monitored_resource.end()) {
+    *base_time_series.mutable_resource() = iter->second;
   } else if (monitored_resource.type().empty()) {
     base_time_series.mutable_resource()->set_type(kDefaultResourceType);
   } else {
--- a/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
+++ b/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
@@ -181,6 +181,8 @@ void SetMetricDescriptor(
 std::vector<google::monitoring::v3::TimeSeries> MakeTimeSeries(
     absl::string_view metric_name_prefix,
     const google::api::MonitoredResource& monitored_resource,
+    const std::unordered_map<std::string, google::api::MonitoredResource>&
+        per_view_monitored_resource,
     const opencensus::stats::ViewDescriptor& view_descriptor,
     const opencensus::stats::ViewData& data,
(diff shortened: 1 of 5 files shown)


opencensus/exporters/stats/stackdriver/internal/stackdriver_utils_test.cc
```diff
@@ -216,7 +222,44 @@ TEST(StackdriverUtilsTest, MakeTimeSeriesCustomResource) {
...
+  const opencensus::stats::ViewData data =
+      TestUtils::MakeViewData(view_descriptor, {{{}, 1.0}});
+  google::api::MonitoredResource resource;
+  resource.set_type("gce_instance");
+  (*resource.mutable_labels())["project_id"] = "my_project";
+  (*resource.mutable_labels())["instance_id"] = "1234";
+  (*resource.mutable_labels())["zone"] = "my_zone";
+  std::unordered_map<std::string, google::api::MonitoredResource>
+      per_view_resources;
+  per_view_resources[view_name] = resource;
```
g-easy: Please also add one entry that shouldn't match. :)

opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
```diff
@@ -181,14 +181,20 @@ void SetMetricDescriptor(
...
+        per_view_monitored_resource,
     const opencensus::stats::ViewDescriptor& view_descriptor,
     const opencensus::stats::ViewData& data,
     absl::string_view opencensus_task) {
   // Set values that are common across all the rows.
   auto base_time_series = google::monitoring::v3::TimeSeries();
   base_time_series.mutable_metric()->set_type(
       MakeType(metric_name_prefix, view_descriptor.name()));
-  if (monitored_resource.type().empty()) {
+  if (per_view_monitored_resource.find(view_descriptor.name()) !=
```
g-easy: This does the search twice. How about something like:
```
auto iter = find(...);
if (iter != end) { set(*iter); } else ...
```

opencensus/exporters/stats/stackdriver/internal/stackdriver_utils_test.cc
bianpengyuan: Done.

opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
bianpengyuan: Ah, yes! Thanks for catching this. Done.

bianpengyuan:
@bogdandrutu Can I get your review on this PR? Thanks!

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
```diff
@@ -48,6 +48,19 @@ struct StackdriverOptions {
...
+  //
+  // If the view name cannot be found in the map, the exporter will use the
+  // "monitored_resource" set in option, or "global" resource if that is not
+  // specified.
+  //
+  // See also:
+  // https://cloud.google.com/monitoring/api/ref_v3/rpc/google.api#google.api.MonitoredResource
+  // https://cloud.google.com/monitoring/api/resources
+  std::unordered_map<std::string, google::api::MonitoredResource>
+      per_view_monitored_resource;
```
bogdandrutu: I would call this per_metric_monitored_resource. When views are applied on the recorded measurements the result is a metric data (I know we call it view_data but that is a small mistake). So I would simply call this per_metric_, also when we will add support for gauge and cumulative metrics this will apply to that as well.

bianpengyuan: Done. Thanks!

PR #363, opened 2019-07-24
Add B3 propagation helpers.

g-easy:


opencensus/trace/internal/b3.cc
```diff
@@ -0,0 +1,93 @@
...
+                          absl::string_view b3_span_id,
+                          absl::string_view b3_sampled) {
+  static SpanContext invalid;
+  uint8_t sampled;
+  if (b3_sampled == "1") {
+    sampled = 1;
+  } else if (b3_sampled == "0" || b3_sampled.empty()) {
+    sampled = 0;
+  } else
+    return invalid;
```
bogdandrutu: I slightly prefer:

```cpp
else {
  return invalid;
}
```

opencensus/trace/internal/b3.cc
```diff
@@ -0,0 +1,93 @@
...
+  if (!IsHexDigits(b3_span_id)) {
+    return invalid;
+  }
+
+  std::string trace_id_binary = absl::HexStringToBytes(b3_trace_id);
+  std::string span_id_binary = absl::HexStringToBytes(b3_span_id);
+
+  // Extend 64-bit trace_id to 128-bit.
+  if (b3_trace_id.length() == 16) {
+    trace_id_binary = std::string(8, '\0') + trace_id_binary;
```
bogdandrutu: I think it is unnecessary to allocate a string every time probably use stack allocated traceid and spanid.

opencensus/trace/propagation/b3.h
```diff
@@ -0,0 +1,63 @@
...
+#include "opencensus/trace/span_context.h"
+
+namespace opencensus {
+namespace trace {
+namespace propagation {
+
+// Implementation of the B3 propagation format:
+// https://github.com/openzipkin/b3-propagation
+
+// We handle:
```
bogdandrutu: We should probably handle https://github.com/openzipkin/b3-propagation#debug-flag as well.

opencensus/trace/internal/b3.cc
g-easy: absl doesn't provide a non-allocating version of HexStringToBytes. I asked and they didn't want to add an API for it.

opencensus/trace/internal/b3.cc
g-easy: Done.

opencensus/trace/propagation/b3.h
g-easy: Done.

opencensus/trace/internal/b3.cc
bogdandrutu: Then at least optimize the 8 zeros string to not allocate. StrCat("00000000", trace_id_binary) maybe make that literal a constat with a static assert to avoid mistakes.

g-easy: I replaced this with:

```
    constexpr char eight_nuls_str[] = "\0\0\0\0\0\0\0\0";
    ABSL_CONST_INIT const absl::string_view eight_nuls(eight_nuls_str, 8);
    trace_id_binary = absl::StrCat(eight_nuls, trace_id_binary);
```

But it's 2ns slower than before.

g-easy: Different approach reduced it from 68 to 44ns.

PR #364, opened 2019-07-24
Stackdriver stats exporter: make opencensus_task optional.

g-easy:
Have the exporter fill it in by default, and move the code out of
examples/ and into opencensus/common/internal/.

### DIFF
--- a/examples/grpc/stackdriver.cc
+++ b/examples/grpc/stackdriver.cc
@@ -12,26 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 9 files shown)


g-easy:
cc @panzhongxian

g-easy:
Oops, need to add the Win32 version.

bogdandrutu:
What is the background of this change?

g-easy:
You wanted to make `opencensus_task` optional / filled in automatically, and the OcAgent exporter is going to need Hostname() so I thought to factor them out like this.

PR #365, opened 2019-07-25
Make delta view available for exporting

### LINKED ISSUE #362: Allow exporting views with Delta aggregation window
Currently view could be registered to use Delta aggregation window:
https://github.com/census-instrumentation/opencensus-cpp/blob/36e10cb7aba4a024f97c8cc4f81a57103ed7baee/opencensus/stats/internal/set_aggregation_window.h#L28

but it cannot be registered to export:
https://github.com/census-instrumentation/opencensus-cpp/blob/36e10cb7aba4a024f97c8cc4f81a57103ed7baee/opencensus/stats/internal/view_descriptor.cc#L81

@g-easy @bogdandrutu  Is there any reason to make it not available for exporting? It would help Stackdriver exporter recovering from timeseries out of order writes issue. Specifically Istio is migrating from one telemetry exporting source to another, which could introduce a brief period with double exporters. WIth cumulative aggregation window, if the old exporter wins the race, the new exporter can never recover from out of order writes unless restart it.

cc @mandarjog @kyessenov

bianpengyuan:
Fixes #362 

@g-easy @bogdandrutu Ptal, thank you!

### DIFF
--- a/opencensus/stats/internal/view_data_impl.cc
+++ b/opencensus/stats/internal/view_data_impl.cc
@@ -278,6 +278,9 @@ ViewDataImpl::ViewDataImpl(ViewDataImpl* source, absl::Time now)
       break;
     }
   }
+  // Add one nanosecond to start time to preventing overlapping the end time of
+  // last interval.
+  start_time_ += absl::Nanoseconds(1);
   source->start_time_ = now;
   source->end_time_ = now;
 }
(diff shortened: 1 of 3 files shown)


bianpengyuan:
cc @kyessenov @mandarjog

g-easy:
Closing per resolution of #362

PR #367, opened 2019-07-26
Build fix: provide a @bazel_version repository.

g-easy:
gRPC depends on upb which depends on bazel_version, which
has to be instantiated in WORKSPACE.

The bazel_version definition is here:

https://github.com/protocolbuffers/upb/blob/1ee1e362565f681fdd46f16bf9f70f51c905ddd4/bazel/repository_defs.bzl


g-easy:
cc @haberman

PR #370, opened 2019-08-06
Switch from custom hash_mix to absl::Hash.

g-easy:
The hash_mix code pre-dates absl::Hash.


g-easy:
No real difference in `//opencensus/tags:tag_map_benchmark` timing.

PR #371, opened 2019-08-06
Build fix: grpc needs build_bazel_apple_support.

g-easy:


g-easy:
I think this dep was added in https://github.com/grpc/grpc/pull/19621
cc @nicolasnoble

gnossen:
Pinging https://github.com/bazelbuild/bazel/issues/1943. Bazel shouldn't be forcing projects to include dependencies like this.

PR #373, opened 2019-08-13
Stackdriver stats: improve handling of builtin metrics.

### LINKED ISSUE #372: Make Stackdriver exporter selectively fill in opecensus_task label
Currently StackdriverOption fill in opecensus_task with a default value (`cpp-{PID}@{HOSTNAME}`) when it is empty:
https://github.com/census-instrumentation/opencensus-cpp/blob/7063ccb607bced1f48eb6a89da1eed162788d783/opencensus/exporters/stats/stackdriver/stackdriver_exporter.h#L35-L38
However when sd exporter is used to export GCP metrics, opencensus_task label should not be filled. We need a way to make opencensus_task label selectively filled for different metrics.

@g-easy @bogdandrutu

g-easy:
* Don't send a CreateMetricDescriptor RPC.
* Don't add the opencensus_task label.

Fixes #372.

### DIFF
--- a/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
+++ b/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
@@ -18,6 +18,7 @@
 
 #include "absl/base/internal/sysinfo.h"
 #include "absl/base/macros.h"
+#include "absl/strings/match.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/string_view.h"
 #include "absl/time/time.h"
@@ -40,11 +41,6 @@ constexpr char kLabelDescription[] = "OpenCensus TagKey";
 constexpr char kOpenCensusTaskKey[] = "opencensus_task";
 constexpr char kOpenCensusTaskDescription[] = "OpenCensus task identifier";
 constexpr char kDefaultResourceType[] = "global";
-
-std::string MakeType(absl::string_view metric_name_prefix,
-                     absl::string_view view_name) {
-  return absl::StrCat(metric_name_prefix, view_name);
-}
 
 // Creates a name in the format described in
 // https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors/create
@@ -153,15 +149,28 @@ std::vector<google::monitoring::v3::TimeSeries> DataToTimeSeries(
 
 }  // namespace
 
+std::string MakeType(absl::string_view metric_name_prefix,
+                     absl::string_view view_name) {
+  return absl::StrCat(metric_name_prefix, view_name);
+}
+
+bool IsBuiltinMetric(absl::string_view metric_type) {
+  return !absl::StartsWith(metric_type, "custom.googleapis.com/") &&
+         !absl::StartsWith(metric_type, "external.googleapis.com/");
+}
+
 void SetMetricDescriptor(
     absl::string_view project_name, absl::string_view metric_name_prefix,
     const opencensus::stats::ViewDescriptor& view_descriptor,
     google::api::MetricDescriptor* metric_descriptor) {
   metric_descriptor->set_name(
       MakeName(project_name, metric_name_prefix, view_descriptor.name()));
-  metric_descriptor->set_type(
-      MakeType(metric_name_prefix, view_descriptor.name()));
-  SetOpenCensusTaskLabelDescriptor(metric_descriptor->add_labels());
+  const std::string metric_type =
+      MakeType(metric_name_prefix, view_descriptor.name());
+  metric_descriptor->set_type(metric_type);
+  if (!IsBuiltinMetric(metric_type)) {
+    SetOpenCensusTaskLabelDescriptor(metric_descriptor->add_labels());
+  }
   for (const auto& tag_key : view_descriptor.columns()) {
     SetLabelDescriptor(tag_key.name(), metric_descriptor->add_labels());
   }
@@ -188,8 +197,9 @@ std::vector<google::monitoring::v3::TimeSeries> MakeTimeSeries(
     absl::string_view opencensus_task) {
   // Set values that are common across all the rows.
   auto base_time_series = google::monitoring::v3::TimeSeries();
-  base_time_series.mutable_metric()->set_type(
-      MakeType(metric_name_prefix, view_descriptor.name()));
+  const std::string metric_type =
+      MakeType(metric_name_prefix, view_descriptor.name());
+  base_time_series.mutable_metric()->set_type(metric_type);
   auto iter = per_metric_monitored_resource.find(view_descriptor.name());
   if (iter != per_metric_monitored_resource.end()) {
     *base_time_series.mutable_resource() = iter->second;
@@ -206,9 +216,10 @@ std::vector<google::monitoring::v3::TimeSeries> MakeTimeSeries(
     SetTimestamp(data.start_time(), interval->mutable_start_time());
   }
   SetTimestamp(data.end_time(), interval->mutable_end_time());
-  (*base_time_series.mutable_metric()->mutable_labels())[kOpenCensusTaskKey] =
-      std::string(opencensus_task);
-
+  if (!IsBuiltinMetric(metric_type)) {
+    (*base_time_series.mutable_metric()->mutable_labels())[kOpenCensusTaskKey] =
+        std::string(opencensus_task);
+  }
   switch (data.type()) {
     case opencensus::stats::ViewData::Type::kDouble:
       return DataToTimeSeries(view_descriptor, data.double_data(),
--- a/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
+++ b/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
@@ -154,9 +154,9 @@ std::string MakeType(absl::string_view metric_name_prefix,
   return absl::StrCat(metric_name_prefix, view_name);
 }
 
-bool IsBuiltinMetric(absl::string_view metric_type) {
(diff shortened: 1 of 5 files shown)


opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
```diff
@@ -63,6 +63,15 @@ struct StackdriverOptions {
...
       per_metric_monitored_resource;
 
+  // Optional: per metric labels to add.
+  //
+  // If the value is empty, the label with the given key will be removed.
+  struct Label {
+    std::string key;
+    std::string value;
+  };
+  std::unordered_map<std::string, Label> per_metric_label;
```
bianpengyuan: This looks great, it could actually avoid lots of copy if some labels remain the same over exporter's lifetime.

One question: when view tag and label map conflict, which one takes precedence?

g-easy: IMO per_metric_label takes precedence, because it's more powerful: you can overwrite and delete columns from your views. (whether doing that is a good idea is a separate discussion)

The comment should be explicit about this.

g-easy: Updated comment.

bianpengyuan:
@bogdandrutu gentle ping, any help on reviewing this? Thank you!

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
```diff
@@ -63,6 +65,20 @@ struct StackdriverOptions {
...
+  // the given key will be removed.
+  struct Label {
+    std::string key;
+    std::string value;
+  };
+  std::unordered_map<std::string, Label> per_metric_label;
+
+  // Optional: metrics for which the exporter should not
+  // CreateMetricDescriptor in Stackdriver.
+  std::unordered_set<std::string> per_metric_do_not_create_descriptor;
```
bogdandrutu: Is this kind of removing?

g-easy: The intent (and I should improve the comment) is to export the metrics, but not send a CreateMetricDescriptor RPC because it will fail because the metric already exists (and we retry failures forever)

g-easy:
@bianpengyuan Just to clarify: do we need `opencensus_task` on some metrics but not others? Or can we add a global "don't add `opencensus_task` to any metrics at all" switch?

g-easy:
Another option: if a metric has a Monitored Resource (e.g. appears in `per_metric_monitored_resource`) then we don't add an `opencensus_task` label to it. Would this address your use-case?

bianpengyuan:
For our use case it could be a global setting.

The other option would address my issue, but I guess you want to attach opencensus_task label to all custom metrics? Custom metric could also have monitored resource specified: https://cloud.google.com/monitoring/custom-metrics/creating-metrics#which-resource.

How about calling [ListMetricDescriptor](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors/list) before calling Create? If the metric is already registered, nothing needs to be done right?

bianpengyuan:
or hard code the logic to only call createMetricDescriptor for metrics prefixed with custom.googleapis.com and external.googleapis.com?

bianpengyuan:
@g-easy Gentle ping, how should we make progress on this? thanks!

bianpengyuan:
Any help on this? I think it is a clean change to hard code the logic to not call `CreateMetricDescriptor` when metric is not prefixed with `custom.googleapis.com` and `external.googleapis.com`. Is there any concern with this approach? If it is reasonable, I could take the implementation work. Thanks!

g-easy:
Sorry for the delay. As I understand it, the features you would like:
* be able to opt out of the opencensus_task label
* be able to opt out of the CreateMetricDescriptor RPC
* to be able to do this on a per-metric basis

Last I looked, there was no precedent for this in opencensus-java.

Bogdan, can we just turn this into:
```c++
set<string> dont_add_opencensus_task;
set<string> dont_create_metric_descriptor;
```

Other options include:

* Suppress opencensus_task label if a metric has a Monitored Resource. Cons: 
  * What if I want to suppress the label but don't have / want a resource.
  * What if I want the label and a resource.

* Only call CreateMetricDescriptor for metrics prefixed with custom.googleapis.com and external.googleapis.com.
  * What happens if that list isn't sufficient? Or is too broad? I don't understand SD prefixes well enough to answer this.

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
bogdandrutu: We have a different logic here in go. It is called `builtinMetric`. See https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/stats.go#L201

Is that enough?

g-easy: That looks like the suggestion in https://github.com/census-instrumentation/opencensus-cpp/pull/373#issuecomment-526382831

@bianpengyuan if we use that logic to suppress both the RPC and the label, is that sufficient?

g-easy:
By default, the exporter uses the prefix `custom.googleapis.com/opencensus/` so we can't use `custom.googleapis.com/` (per the golang version) to suppress the CreateMetricDescriptor RPC. Right?

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
bianpengyuan: Yes, that would be sufficient.

bianpengyuan:
@g-easy Is it possible to make it: if prefix is not given or is not prefixed by `external.googleapis.com`?

g-easy:
Sure. How about this IsBuiltinMetric change?

opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
```diff
@@ -153,15 +149,27 @@ std::vector<google::monitoring::v3::TimeSeries> DataToTimeSeries(
...
 void SetMetricDescriptor(
     absl::string_view project_name, absl::string_view metric_name_prefix,
     const opencensus::stats::ViewDescriptor& view_descriptor,
     google::api::MetricDescriptor* metric_descriptor) {
   metric_descriptor->set_name(
       MakeName(project_name, metric_name_prefix, view_descriptor.name()));
-  metric_descriptor->set_type(
-      MakeType(metric_name_prefix, view_descriptor.name()));
-  SetOpenCensusTaskLabelDescriptor(metric_descriptor->add_labels());
+  const std::string metric_type =
```
bianpengyuan: Can we assume this function will only be triggered for non-builtin metric?

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
g-easy: Going with the `IsBuiltinMetric()` approach. Resolved.

opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
g-easy: I'd like to leave this code as-is.

g-easy:
PTAL.

opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
```diff
@@ -153,15 +149,28 @@ std::vector<google::monitoring::v3::TimeSeries> DataToTimeSeries(
 
 }  // namespace
 
+std::string MakeType(absl::string_view metric_name_prefix,
+                     absl::string_view view_name) {
+  return absl::StrCat(metric_name_prefix, view_name);
+}
+
+bool IsBuiltinMetric(absl::string_view metric_type) {
```
helly25: honestly this looks very strange since this is a negative check against known custom metrics. That is not the opposite of builtin metrics, since there is no knowledge of identifying those.

g-easy: I also think it's a bit odd but it's what the golang version does. It's documented as "heuristically known"

helly25: A repeated failure doesn't make a good feature. That said, this could at least be named: IsKnownCustomMetric.

bogdandrutu:
LGTM - I like @helly25 proposal about the name.

opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
g-easy: Done.

opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc
```diff
@@ -153,15 +149,28 @@ std::vector<google::monitoring::v3::TimeSeries> DataToTimeSeries(
...
     google::api::MetricDescriptor* metric_descriptor) {
   metric_descriptor->set_name(
       MakeName(project_name, metric_name_prefix, view_descriptor.name()));
-  metric_descriptor->set_type(
-      MakeType(metric_name_prefix, view_descriptor.name()));
-  SetOpenCensusTaskLabelDescriptor(metric_descriptor->add_labels());
+  const std::string metric_type =
+      MakeType(metric_name_prefix, view_descriptor.name());
+  metric_descriptor->set_type(metric_type);
+  if (IsKnownCustomMetric(metric_type)) {
```
bogdandrutu: Probably add a comment to say why.

g-easy: https://github.com/census-instrumentation/opencensus-cpp/pull/385

PR #376, opened 2019-08-13
Make B3 parser more readable.

g-easy:
No functional change.

### DIFF
--- a/opencensus/trace/internal/b3.cc
+++ b/opencensus/trace/internal/b3.cc
@@ -75,15 +75,19 @@ SpanContext FromB3Headers(absl::string_view b3_trace_id,
   std::string trace_id_binary = absl::HexStringToBytes(b3_trace_id);
   std::string span_id_binary = absl::HexStringToBytes(b3_span_id);
 
-  // trace_id_ptr must point to a 128-bit trace_id.
-  const uint8_t* trace_id_ptr =
-      reinterpret_cast<const uint8_t*>(trace_id_binary.data());
   uint8_t extended_trace_id[16];
-  if (trace_id_binary.length() != 16) {
+
+  // trace_id_ptr must point to a 128-bit trace_id.
+  const uint8_t* trace_id_ptr;
+  if (trace_id_binary.length() == 16) {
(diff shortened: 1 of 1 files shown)


opencensus/trace/internal/b3.cc
```diff
@@ -75,17 +75,19 @@ SpanContext FromB3Headers(absl::string_view b3_trace_id,
   std::string trace_id_binary = absl::HexStringToBytes(b3_trace_id);
   std::string span_id_binary = absl::HexStringToBytes(b3_span_id);
 
-  // Extend 64-bit trace_id to 128-bit.
+  // trace_id_ptr must point to a 128-bit trace_id.
+  const uint8_t* trace_id_ptr =
+      reinterpret_cast<const uint8_t*>(trace_id_binary.data());
   uint8_t extended_trace_id[16];
-  if (b3_trace_id.length() == 16) {
+  if (trace_id_binary.length() != 16) {
```
helly25: Switching to != 16 is the better option, but now you have an issue because there is zero guarantee that != 16 means size=8. Please make this code resilient to changing sizes.

g-easy: L65 guarantees it'll be either 16 or 8 after HexStringToBytes. I'll try duplicating the check here and let's see if it'll be more readable.

g-easy:
Benchmark before:
```
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
BM_FromB3Headers_128bitTraceId        76.7 ns         76.7 ns      6863265
BM_FromB3Headers_64bitTraceId         45.2 ns         45.2 ns     15227647
BM_FromB3Headers_InvalidTraceId       25.1 ns         25.1 ns     30997309
```

After:
```
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
BM_FromB3Headers_128bitTraceId        74.4 ns         74.4 ns      8511255
BM_FromB3Headers_64bitTraceId         43.1 ns         43.1 ns     16243200
BM_FromB3Headers_InvalidTraceId       29.0 ns         29.0 ns     23450057
```

PR #378, opened 2019-08-21
Build fix.

g-easy:
Need rules_cc for abseil and build_bazel_rules_apple for grpc.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -16,6 +16,13 @@ workspace(name = "io_opencensus_cpp")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+# Build rules for C++ projects.
+http_archive(
+    name = "rules_cc",
+    strip_prefix = "rules_cc-master",
+    urls = ["https://github.com/bazelbuild/rules_cc/archive/master.zip"],
+)
+
 # We depend on Abseil.
 http_archive(
     name = "com_google_absl",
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -16,6 +16,13 @@ workspace(name = "io_opencensus_cpp")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+# Build rules for C++ projects.
+http_archive(
```
gnossen: For what it's worth, I would consider the fact that this dependency had to be added to your WORKSPACE a bug in the `abseil-cpp` repo. There doesn't appear to be anything stopping them from pulling this out to a `deps()` macro.

WORKSPACE
```diff
@@ -39,6 +46,20 @@ http_archive(
     urls = ["https://github.com/google/benchmark/archive/master.zip"],
 )
 
+# Used by gRPC.
+http_archive(
```
gnossen: You shouldn't need this particular `http_archive` invocation unless for some reason you really want `master` branch (which, btw, makes your build non-hermetic). You're already getting an instance from your call to `grpc_deps()`. It's just the `load()` and `apple_rules_dependencies()` invocations below that are necessary due to https://github.com/bazelbuild/bazel/issues/1550.

WORKSPACE
g-easy: Please correct me if I'm wrong but my understanding is that this will be required for all C++ projects after bazel 1.0(?) because rules like `cc_library` will move from native to skylark implementations.

WORKSPACE
g-easy: Thanks! How does https://github.com/census-instrumentation/opencensus-cpp/pull/379 look to you?

As for master, we are trying to follow Abseil's "live at head" philosophy. (...on our own master branch; whereas OpenCensus releases pin their deps)

WORKSPACE
gnossen: That may be the case for one hop. But suppose absl CPP requires `rules_cc` and your repo provides, say, a Golang tool. It's not intuitive to users of the Golang tool that they should have to pull in `rules_cc` despite never directly using a C++ rule.

This has already happened in the gRPC repo. We have to pull in a Golang-only tool used by Envoy even though we're only -- to my knowledge -- using some of their protos.

g-easy: Ditto. opencensus-cpp pulls in e.g. grpc-java, because opencensus-proto depends on it.

As for rules_cc specifically, I think opencensus-cpp will end up using them directly after bazel 1.0.

PR #380, opened 2019-08-23
Simplify WORKSPACE file.

panzhongxian:
I found the WORKSPACE of opencensus-cpp-example project too old to work.

I split the repositories from WORKSPACE.  A simple WORKSPACE may helpful for the library user.

I also want to  move the rest `load` and dependence call out. so that we can only use simplely in both opencensus-cpp and other projects :
```
workspace(name = "io_opencensus_cpp")

load(":deps.bzl", "opencensus_cpp_deps")

opencensus_cpp_deps()
```

Do you have any idea?


WORKSPACE
```diff
@@ -14,170 +14,30 @@
 
 workspace(name = "io_opencensus_cpp")
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load(":deps.bzl", "opencensus_cpp_deps")
```
g-easy: IMO we should put deps.bzl into the bazel/ directory to match what other projects (grpc, upb, envoy, google-cloud-cpp) do.

WORKSPACE
```diff
@@ -14,170 +14,30 @@
...
-http_archive(
-    name = "com_google_absl",
-    strip_prefix = "abseil-cpp-master",
-    urls = ["https://github.com/abseil/abseil-cpp/archive/master.zip"],
-)
-
-# GoogleTest framework.
-# Only needed for tests, not to build the OpenCensus library.
-http_archive(
-    name = "com_google_googletest",
```
g-easy: googletest is not used for building, only for testing, so IMO it should be in WORKSPACE directly, and not part of `opencensus_cpp_deps()`

Ditto benchmark.

deps.bzl
```diff
@@ -0,0 +1,179 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+def opencensus_cpp_deps():
+    """Loads dependencies need to compile and test the opencensus-cpp library."""
+
+    # native.bind():
+
+    if "com_google_protobuf" not in native.existing_rules():
```
g-easy: I think this can be done via `maybe()`

opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
```diff
@@ -38,8 +38,9 @@ namespace {
 // Prometheus's name requirements.
 std::string SanitizeName(absl::string_view name) {
   std::string sanitized(name);
-  std::replace_if(sanitized.begin(), sanitized.end(),
-                  [](char c) { return !::isalnum(c); }, '_');
+  std::replace_if(
+      sanitized.begin(), sanitized.end(), [](char c) { return !::isalnum(c); },
```
g-easy: Did clang-format do this?

deps.bzl
```diff
@@ -0,0 +1,179 @@
...
+            urls = ["https://github.com/jupp0r/prometheus-cpp/archive/master.zip"],
+        )
+
+    if "com_github_curl" not in native.existing_rules():
+        # Curl library - used by zipkin exporter.
+        http_archive(
+            name = "com_github_curl",
+            build_file_content =
+                """
+        load("@io_opencensus_cpp//opencensus:curl.bzl", "CURL_COPTS")
```
g-easy: Please move this block of text into a new file, bazel/curl.BUILD

deps.bzl
```diff
@@ -0,0 +1,179 @@
...
+            strip_prefix = "curl-master",
+            urls = ["https://github.com/curl/curl/archive/master.zip"],
+        )
+
+    if "com_github_tencent_rapidjson" not in native.existing_rules():
+        # Rapidjson library - used by zipkin exporter.
+        http_archive(
+            name = "com_github_tencent_rapidjson",
+            build_file_content =
+                """
```
g-easy: bazel/rapidjson.BUILD

panzhongxian:
We could also create one function per dep like prometheus does: https://github.com/jupp0r/prometheus-cpp/blob/master/repositories.bzl#L46
But as g-easy says, the deps are loaded on demand, so it's simpler to have one big function. Also that's how grpc does it.

panzhongxian:
ERROR: /home/travis/build/census-instrumentation/opencensus-cpp/bazel/deps.bzl:2:1: file '@bazel_tools//tools/build_defs/repo:utils.bzl' does not contain symbol 'maybe'

@g-easy  Can you help to update bazel version on travis-ci?

opencensus/exporters/stats/prometheus/internal/prometheus_utils.cc
panzhongxian: Yes. Should I recover it?

g-easy:
Please update `BAZEL_VERSION` in `tools/travis/build_bazel.sh`

bazel/deps.bzl
```diff
@@ -0,0 +1,105 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+
+def opencensus_cpp_deps():
+    """Loads dependencies need to compile and test the opencensus-cpp library."""
```
g-easy: Not test.

bazel/deps.bzl
```diff
@@ -0,0 +1,105 @@
...
+        http_archive,
+        name = "com_google_absl",
+        strip_prefix = "abseil-cpp-master",
+        urls = ["https://github.com/abseil/abseil-cpp/archive/master.zip"],
+    )
+
+    # # Used by gRPC.
+    # maybe(
+    #     http_archive,
+    #     name = "build_bazel_rules_apple",
```
g-easy: Please remove rules_apple and apple_support, we get these via grpc.

bazel/curl.BUILD
```diff
@@ -0,0 +1,32 @@
+load("@io_opencensus_cpp//opencensus:curl.bzl", "CURL_COPTS")
+
```
g-easy: Please format this file with buildifier.

WORKSPACE
```diff
@@ -172,25 +78,8 @@ switched_rules_by_language(
     grpc = True,
 )
 
-# Needed by @opencensus_proto.
```
g-easy: Please keep this comment line (for the load() below)

WORKSPACE
```diff
@@ -83,87 +65,11 @@ local_repository(
...
-    ]),
-    includes = ["include/"],
-    visibility = ["//visibility:public"],
-)
-""",
-    strip_prefix = "rapidjson-master",
-    urls = ["https://github.com/Tencent/rapidjson/archive/master.zip"],
-)
-
-# Google APIs - used by Stackdriver exporter.
```
g-easy: Please keep this comment line for the load() below.

bazel/curl.BUILD
```diff
@@ -0,0 +1,32 @@
...
+    srcs = glob([
+        "lib/**/*.c",
+    ]),
+    hdrs = glob([
+        "include/curl/*.h",
+        "lib/**/*.h",
+    ]),
+    includes = ["include/", "lib/"],
+    copts = CURL_COPTS + [
+        "-DOS=\"os\"",
```
g-easy: "-DOS=os",

panzhongxian: It should be `\"os\"`.
If I change this to `os`, there is an error:
```
external/com_github_curl/lib/version.c:302:3: error: use of undeclared identifier 'os'
  OS, /* as found by configure or set by hand at build-time */
  ^
<command line>:162:12: note: expanded from here
#define OS os
```

g-easy: Ooops! I'm wrong. Please change it back.

bazel/deps.bzl
```diff
@@ -27,14 +27,6 @@ def opencensus_cpp_deps():
...
-    # maybe(
-    #     http_archive,
-    #     name = "build_bazel_rules_apple",
-    #     strip_prefix = "rules_apple-master",
-    #     urls = ["https://github.com/bazelbuild/rules_apple/archive/master.zip"],
-    # )
-
     # Used by gRPC.
     maybe(
         http_archive,
```
g-easy: Please also delete this one.

bazel/deps.bzl
```diff
@@ -0,0 +1,89 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+
+def opencensus_cpp_deps():
+    """Loads dependencies need to compile the opencensus-cpp library."""
+
+    maybe(
+        http_archive,
+        name = "com_google_protobuf",
```
g-easy: Where did this come from? The current `WORKSPACE` doesn't load `com_google_protobuf`.

bazel/deps.bzl
g-easy: (I did this in an edit)

bazel/deps.bzl
panzhongxian: I pasted it here maybe when I try to use maybe function.  😿 
Deleted.

PR #381, opened 2019-08-23
Rename io_opencensus_proto to opencensus_proto to match Envoy.

g-easy:


g-easy:
The opencensus-proto WORKSPACE file implies that the repo name should be `opencensus_proto` without the `io_*` prefix. I opened https://github.com/census-instrumentation/opencensus-proto/issues/208 asking if this is intentional.

PR #384, opened 2019-08-29
Add default cases for view data switch

zachary-trudo:
View data switch without default causes compiler warnings.
 Addition of a default removes these warnings.


g-easy:
Thanks for the cleanup! Related issue: https://github.com/census-instrumentation/opencensus-cpp/issues/79

PR #389, opened 2019-09-17
Stackdriver stats: resource must uniquely identify a timeseries.

g-easy:
- Don't add the opencensus_task label when a non-default
MonitoredResource is present.

- Refactor exporter code to reduce duplication of effort.

### DIFF
--- a/opencensus/exporters/stats/stackdriver/internal/stackdriver_exporter.cc
+++ b/opencensus/exporters/stats/stackdriver/internal/stackdriver_exporter.cc
@@ -61,8 +61,8 @@ class Handler : public ::opencensus::stats::StatsExporter::Handler {
   // Returns true if the view has already been registered or registration is
   // successful, and false if the registration fails or the name has already
   // been registered with different parameters.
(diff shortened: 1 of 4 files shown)


opencensus/exporters/stats/stackdriver/internal/stackdriver_exporter.cc
```diff
@@ -103,13 +103,34 @@ void Handler::ExportViewData(
...
+
+    const bool is_known_custom_metric = IsKnownCustomMetric(metric_type);
+
+    // If there is a non-default MonitoredResource for this view, it must
+    // uniquely identify the timeseries.
+    //
+    // Otherwise, we add the opencensus_task label so that different processes
+    // produce different timeseries instead of colliding.
+    const bool add_task_label =
+        is_known_custom_metric && (monitored_resource_for_view == nullptr);
```
helly25: Is && here actually correct - I cannot quite follow the explanation.

g-easy: Updated the comment.

PR #392, opened 2019-09-24
Allow build with other than C++11 standard

bitomaxsp:


g-easy:
Can we do something like: set CMAKE_CXX_STANDARD to 11 unless it's already set

(by an outside build that's pulling in opencensus-cpp via FetchProject_Add?)

bitomaxsp:
@g-easy That was what I initially did and it does not work. I did override to CXX17. I use `fetchcontent_declare` but opencensus project is built with CXX11.

I was quite surprised that is does not have override because prometeus-cpp has it for exmaple.

g-easy:
Why doesn't it work? Can you show me the outer build?

I'm not sure Prometheus's approach is the right one: https://github.com/jupp0r/prometheus-cpp/blob/a0d7c52ac4b48f4f2bb32f5c6f452205594afe1a/CMakeLists.txt#L16

bitomaxsp:
@g-easy Actually I realized that I misinterpreted your suggestion and tried something different. I'll try now what you proposed

bitomaxsp:
@g-easy I've changed the PR. It seem to work nice. Thank you for suggestion.

g-easy:
I like this change. If this solves your problem and Travis passes, I'm happy to merge this.

PR #395, opened 2019-10-02
When getting ViewData, set the end_time to the current time.

g-easy:
Exporters need to export the current timestamp to establish that
a value still holds, not only the last time that it changed.

### DIFF
--- a/opencensus/stats/internal/view_data_impl.h
+++ b/opencensus/stats/internal/view_data_impl.h
@@ -56,7 +56,6 @@ class ViewDataImpl {
   // snapshotting cumulative data; ViewDataImpls for interval views must be
   // converted using the following constructor before snapshotting.
   ViewDataImpl(absl::Time start_time, const ViewDescriptor& descriptor);
-
   // Constructs a ViewDataImpl capturing the state of 'other' at 'now'. Requires
   // 'other' to have an interval aggregation window (and thus type()
   // kStatsObject).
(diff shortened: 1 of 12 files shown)


opencensus/stats/view.h
```diff
@@ -48,8 +50,8 @@ class View {
   // Returns true if this object is valid and data can be collected.
   bool IsValid() const;
 
-  // Returns a snapshot of the View's data.
-  const ViewData GetData();
+  // Returns a snapshot of the View's data with the given time.
+  const ViewData GetData(absl::Time now = absl::Now());
```
bogdandrutu: Do we need the parameter of the time here? I think we should always use Now() because the value represents the value until now.

opencensus/stats/internal/view_data_impl.h
```diff
@@ -106,6 +107,9 @@ class ViewDataImpl {
   absl::Time start_time() const { return start_time_; }
   absl::Time end_time() const { return end_time_; }
```
bogdandrutu: Can we remove end_time from the view_data_impl? End time is always the moment when we read the data, because we read all the updates until that moment (I know it is an approximation, because we do have some async logic, but it is good enough).

g-easy: We would have to also remove end_time from the public API in `opencensus/stats/view_data.h`

```c++
absl::Time ViewData::end_time() const { return impl_->end_time(); }
```

opencensus/stats/view.h
g-easy: I wanted it both for consistency and as an optimization: call `absl::Now()` once in the `StatsExporter` and use the same time for all views.

opencensus/stats/internal/view_data_impl.h
bogdandrutu: That will be the value from the getdata?

g-easy: Ok, so do you mean:
* Keep the public API `ViewData::end_time` and
* Replace internal `ViewDataImpl::end_time` with a new member `ViewData::end_time_`

opencensus/stats/view.h
g-easy: Reverted this.

PR #402, opened 2019-10-29
Speed up SetTimestamp benchmark.

g-easy:
Move the proto constructor out of the loop.


g-easy:
```
Benchmark                 Time             CPU   Iterations
-----------------------------------------------------------
BM_SetTimestamp        6.64 ns         6.64 ns    105402846 before
BM_SetTimestamp        4.21 ns         4.21 ns    166601880 after
```

PR #407, opened 2019-11-01
gRPC example: add custom tags in client.

g-easy:


examples/grpc/hello_client.cc
```diff
@@ -65,8 +68,14 @@ int main(int argc, char **argv) {
       "HelloClient", /*parent=*/nullptr, {&sampler});
   std::cout << "HelloClient span context is " << span.context().ToString()
             << "\n";
+
+  // Create a tag map.
+  static const auto key = opencensus::tags::TagKey::Register("my_key");
+  opencensus::tags::TagMap tags({{key, "my_value"}});
```
bogdandrutu: This should extend the current TagMap correct? That's usually the case.

g-easy: Good point. Added. PTAL.

PR #410, opened 2019-11-05
[Stackdriver Exporter] Avoid sending time series that is 24h in the past

bianpengyuan:
Reference: https://github.com/istio/istio/issues/18613

Not sure if stale view data (time series) is expected, maybe we need to track debugging of that. This is to put a quick bandaid to avoid sending data that the backend does not want.


bianpengyuan:
cc @douglas-reid @mandarjog

bogdandrutu:
I think this was caused by this https://github.com/census-instrumentation/opencensus-cpp/pull/395

bianpengyuan:
Yeah, envoyproxy has not been updated to include the #395. I will update envoyproxy's sha instead.

PR #413, opened 2019-11-18
gRPC example: show current span context and tags in server.

g-easy:


g-easy:
BTW I've fixed the build break in googleapis. I'm waiting for that change to show up in their repo. :)

PR #419, opened 2019-11-28
Clean up bazel cache after build.

g-easy:
Otherwise it grows without bound.


g-easy:
This isn't finding any files to delete. I suspect the way Travis manages its cache, it doesn't restore the correct atime on files.

PR #422, opened 2019-11-28
Make stats export interval configurable at runtime.

### LINKED ISSUE #412: Make stats export interval configurable
Right now export interval is set to be 10s by default and is not configurable. However each backend has its own recommended minimum exporting period, i.e. stackdriver recommends minimum exporting period is 1 min. Opencensus go library supports configurable export interval: https://github.com/census-instrumentation/opencensus-go/blob/df6e2001952312404b06f5f6f03fcb4aec1648e5/stats/view/worker.go#L119 , we should consider adding it here as well.

@g-easy @bogdandrutu 

FYI @quentinmit @mandarjog

g-easy:
Fixes #412.

### DIFF
--- a/opencensus/stats/internal/stats_exporter.cc
+++ b/opencensus/stats/internal/stats_exporter.cc
@@ -37,6 +37,12 @@ StatsExporterImpl* StatsExporterImpl::Get() {
   return global_stats_exporter_impl;
 }
 
+// static
+void StatsExporterImpl::SetInterval(absl::Duration interval) {
+  absl::MutexLock l(&mu_);
+  export_interval_ = interval;
+}
+
 void StatsExporterImpl::AddView(const ViewDescriptor& view) {
   absl::MutexLock l(&mu_);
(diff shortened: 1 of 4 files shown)


opencensus/stats/internal/stats_exporter.cc
```diff
@@ -90,31 +96,51 @@ void StatsExporterImpl::StartExportThread() EXCLUSIVE_LOCKS_REQUIRED(mu_) {
 }
 
 void StatsExporterImpl::RunWorkerLoop() {
-  absl::Time next_export_time = absl::Now() + export_interval_;
+  absl::Time next_export_time;
+  {
+    absl::MutexLock l(&mu_);
+    next_export_time = absl::Now() + export_interval_;
```
bogdandrutu: Make this block a function? To avoid repeating?

g-easy: Done.

PR #426, opened 2019-12-03
Use bazel --curses=no to reduce the amount of bazel chattiness.

g-easy:


g-easy:
You can see the new look [here](https://travis-ci.org/census-instrumentation/opencensus-cpp/jobs/619905494).

PR #427, opened 2019-12-20
Simplify bazel test filtering.

g-easy:


g-easy:
Build should be fixed by #428.

PR #438, opened 2020-03-24
Travis: upgrade the OS image from trusty to xenial.

g-easy:
Trusty is EOL. Xenial is default since 2019-08-13.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,23 +4,23 @@ compiler: clang
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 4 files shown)


g-easy:
This was motivated by fixing the build due to prometheus-cpp requiring a newer CMake version than trusty has: https://github.com/jupp0r/prometheus-cpp/commit/c06e0ef4e46a4770faa10210d41974df1fa12941

PR #440, opened 2020-05-05
Update the Stackdriver exporter to not user overlapping start time an…

ajamato:
Update the Stackdriver exporter to not user overlapping start time and end time timestamps. Doing so can cause stackdriver cumulative metrics to clear the previous overlapping window


ajamato:
@g-easy PTAL

PR #441, opened 2020-05-05
Update the Stackdriver exporter to not user overlapping timestamps

ajamato:
Update the Stackdriver exporter to not user overlapping start time and end time timestamps. Doing so can cause stackdriver cumulative metrics to clear the previous overlapping window


ajamato:
Not sure if this is the correct change, let's not merge this quite yet. Until I debug a bit more please.

PR #442, opened 2020-05-15
Move and rename the stackdriver exporter Handler to header, add virtual PrepareClientContext

ajamato:
Move and rename the stackdriver exporter Handler to the header. Add a method which allows a subclass to apply a custom logic to modify the GrpcClientContext. I.e. to add CallCredentials

### DIFF
--- a/opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
+++ b/opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
@@ -80,6 +80,14 @@ struct StackdriverOptions {
   // instead. Useful for testing.
   std::unique_ptr<google::monitoring::v3::MetricService::StubInterface>
       metric_service_stub;
+
+  // Optional: A function which can be used to modify the grpc::ClientContext
+  // before making API requests. For example, to add custom credentials.
+  // This function will be called by the exporter before every CreateTimeSeries
+  // and CreateMetricDescriptor RPC.
+  // This function will not be called in parallel.
+  std::function<void(grpc::ClientContext*)> prepare_client_context =
+      [](grpc::ClientContext*) {};
 };
 
 // Exports stats for registered views (see opencensus/stats/stats_exporter.h) to
--- a/opencensus/exporters/stats/stackdriver/internal/stackdriver_exporter.cc
+++ b/opencensus/exporters/stats/stackdriver/internal/stackdriver_exporter.cc
@@ -14,13 +14,14 @@
 
 #include "opencensus/exporters/stats/stackdriver/stackdriver_exporter.h"
 
+#include <grpcpp/grpcpp.h>
+
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
 #include <memory>
 #include <vector>
 
-#include <grpcpp/grpcpp.h>
 #include "absl/memory/memory.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/string_view.h"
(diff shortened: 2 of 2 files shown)


opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
```diff
@@ -107,6 +108,37 @@ class StackdriverExporter {
   StackdriverExporter() = delete;
 };
 
+class StackdriverExporterHandler
```
g-easy, line 111: Our intent for this API was that users only need to call `StackdriverExporter::Register(std::move(opts));`

How does subclassing the handler change this? Should we instead have a callback in the options?

ajamato: Okay, I added a std::function to the options class. Makes it a much simpler PR overall, only a few lines change. :). This works for me.

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
```diff
@@ -80,6 +80,12 @@ struct StackdriverOptions {
   // instead. Useful for testing.
   std::unique_ptr<google::monitoring::v3::MetricService::StubInterface>
       metric_service_stub;
+
+  // Optional: A function which can be used to modify the grpc::ClientContext
+  // before making API requests.
+  // This can be used to add custom credentials for example.
+  std::function<void(grpc::ClientContext*)> prepare_client_context =
+      std::function<void(grpc::ClientContext*)>();
```
g-easy, line 88: Is the explicit initialization needed? I thought std::function by default initialized to something nullptr-like.

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
```diff
@@ -80,6 +80,12 @@ struct StackdriverOptions {
   // instead. Useful for testing.
   std::unique_ptr<google::monitoring::v3::MetricService::StubInterface>
       metric_service_stub;
+
+  // Optional: A function which can be used to modify the grpc::ClientContext
+  // before making API requests.
+  // This can be used to add custom credentials for example.
```
g-easy, line 86: Please add:
This function will be called by the exporter before every CreateTimeSeries and CreateMetricDescriptor RPC.
This function will not be called in parallel.

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
ajamato: If its not set it will throw an exception when called (which I tested and verified is true):
https://en.cppreference.com/w/cpp/utility/functional/function

If we don't set a default, we would need to use it with a check first, like this:
if (prepare_client_context) {
  prepare_client_context(context);
}

The stored callable object is called the target of std::function. If a std::function contains no target, it is called empty. Invoking the target of an empty std::function results in std::bad_function_call exception being thrown.

Note: I realized that the bazel test targets don't test this. But I was able to confirm the std::function being called when I ran the example which exports to stackdriver. I had to change the syntax to the lambda expression

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
ajamato: Done

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
g-easy: Thanks!

opencensus/exporters/stats/stackdriver/stackdriver_exporter.h
g-easy: I was wrong about the null-like. Thanks for following up.

PR #445, opened 2020-05-21
Stackdriver trace exporter: add prepare_client_context callback.

g-easy:


g-easy:
#442 took care of the stats exporter, this PR is for the trace exporter.

PR #449, opened 2020-06-20
updates to tools/docker-format/Dockerfile : container builds again

alichnewsky:
current docker file fails to build.
As a result, can't run pre-commit formatting before PR fail in CI builds.

to validate rebuild the container as per instructions in the README.md file.

### DIFF
--- a/tools/docker-format/Dockerfile
+++ b/tools/docker-format/Dockerfile
@@ -1,8 +1,28 @@
 FROM ubuntu:cosmic
 
+# repositories are obsolete. the image need to be changed.
+RUN perl -pi -e 's@//archive@//old-releases@g' /etc/apt/sources.list
+# disable security updates altogether
+RUN perl -pi -e 's@^(.*security.*)$@#$1@g' /etc/apt/sources.list
+
 RUN apt update && \
-    apt install -y clang-format golang git python-pip && \
-    go get -v github.com/bazelbuild/buildtools/buildifier && \
+    apt install -y clang-format git python-pip && \
     pip install 'cmake_format>=0.5.2'
+
+# brings golang to 1.10 by default.bazelbuilds wants 1.12 minimum....
+#RUN apt install -y golang
+
+RUN apt install -y curl
+RUN curl -sL https://dl.google.com/go/go1.13.3.linux-amd64.tar.gz -o go1.13.3.linux-amd64.tar.gz
+RUN tar -xf go1.13.3.linux-amd64.tar.gz
+RUN mv go /usr/local/go-1.13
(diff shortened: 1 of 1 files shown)


tools/docker-format/Dockerfile
```diff
@@ -1,8 +1,28 @@
 FROM ubuntu:cosmic
 
+# repositories are obsolete. the image need to be changed.
```
g-easy, line 3: If we update to something newer than cosmic, do we also get golang>=1.12?

alichnewsky: Upgrading from cosmic to focal ( 20.04 LTS ) gives you exactly golang 1.13. 
It requires python3 but the same cmake-format package version works. 
The default clang-format is 10.0 and needs to be downgraded not to have side effects on the current c++ code.

On other hand, installing golang manually makes the upgrade path clear should buildifier requirements change again.

currently in this PR
```
docker run -v $PWD:/opencensus-cpp -it opencensus-cpp/format:cosmic /opencensus-cpp/tools/docker-format/run.sh
clang-format version 7.0.0-3 (tags/RELEASE_700/final)
Running buildifier.
Running cmake-format 0.6.10.
```

potential "focal" version
```
docker run -v $PWD:/opencensus-cpp -it opencensus-cpp/format:focal /opencensus-cpp/tools/docker-format/run.sh

clang-format version 7.0.1-12 (tags/RELEASE_701/final)
Running buildifier.
Running cmake-format 0.6.10.
```

I can submit the latter if you'd rather.

g-easy: 1. Is the "focal" version simpler?

2. Instead of installing golang in "cosmic" is it easier to just install a release of buildifier? e.g. https://github.com/bazelbuild/buildtools/releases/download/3.3.0/buildifier

tools/docker-format/Dockerfile
```diff
@@ -1,8 +1,28 @@
...
 RUN apt update && \
-    apt install -y clang-format golang git python-pip && \
-    go get -v github.com/bazelbuild/buildtools/buildifier && \
+    apt install -y clang-format git python-pip && \
     pip install 'cmake_format>=0.5.2'
 
+# brings golang to 1.10 by default.bazelbuilds wants 1.12 minimum....
+#RUN apt install -y golang
+
+RUN apt install -y curl
```
g-easy, line 15: Please merge all of the RUNs together into a single one to avoid a long chain of docker overlays.

lgruen (changes requested):
Sorry, I'm not working on OpenCensus anymore, but can't remove myself as a reviewer from this PR. Dismissing this review using "request changes" might remove it from my PR queue though.

Issue #451, opened 2020-08-21
Metrics should expire from stackdriver time_series/view vector


mandarjog:
The following vector grows indefinitely in memory.

https://github.com/census-instrumentation/opencensus-cpp/blob/85e29c4d815171656d54fdf2dc5b74873a4a62bd/opencensus/exporters/stats/stackdriver/internal/stackdriver_exporter.cc#L106

1. This vector should be pruned periodically.
2. It should be possible to access the current metrics cardinality, or present it during pruning.

g-easy:
IIUC, pruning is motivated by wanting to have a "sparse" map where timeseries eventually disappear. What is the logic for when a timeseries should reset / disappear? Or should the stackdriver exporter expose an API for "reset some timeseries" and have the calling code specify the logic?

mandarjog:
We can start simple by exposing a `clear()` api which should `clear()` state at `next export` so as to not drop any data.
@bianpengyuan ?

bianpengyuan:
@g-easy any feedback on this approach? Do you think if this is feasible without major surgery to the code?

bianpengyuan:
Talked with mandarjog, the other option is to add a SetExpiry() method into StatsManager, and whenever a harvest cycle happens, clean up view that has been updated for the given period. Comparing to the Clean() method, this would help to flatten the latency from lock.

g-easy:
Expiry done as part of the existing background thread stats aggregation sounds reasonable to me. Should expiry duration be part of StatsManager, or part of the ViewDescriptor?

bianpengyuan:
Ah, yes, viewDescriptor makes more sense. I can take a stab on this.

g-easy:
Yes please. I'm happy to review but I'm very limited in how much time I can spend on OC these days. :( 

BTW I suggest ViewDescriptor because it's part of the public API whereas StatsManager is not.

PR #452, opened 2020-09-13
Add view data expiry

bianpengyuan:
fixes https://github.com/census-instrumentation/opencensus-cpp/issues/451

### DIFF
--- a/opencensus/stats/internal/view_data_impl.cc
+++ b/opencensus/stats/internal/view_data_impl.cc
@@ -197,7 +197,7 @@ void ViewDataImpl::Merge(const std::vector<std::string>& tag_values,
   // A value is set here. Set a start time if it is unset.
   SetStartTimeIfUnset(tag_values, now);
   SetUpdateTime(tag_values, now);
-  RemoveExpired(now);
+  PurgeExpired(now);
   switch (type_) {
     case Type::kDouble: {
       if (aggregation_.type() == Aggregation::Type::kSum) {
@@ -327,7 +327,7 @@ void ViewDataImpl::SetUpdateTime(const std::vector<std::string>& tag_values,
   }
 }
 
-void ViewDataImpl::RemoveExpired(absl::Time now) {
+void ViewDataImpl::PurgeExpired(absl::Time now) {
   if (expiry_duration_ == absl::ZeroDuration() || update_times_.empty()) {
     // No need to remove expired entries since either expiry is not set or there
     // is no data entry yet.
--- a/opencensus/stats/internal/view_data_impl.cc
+++ b/opencensus/stats/internal/view_data_impl.cc
@@ -197,6 +197,7 @@ void ViewDataImpl::Merge(const std::vector<std::string>& tag_values,
   // A value is set here. Set a start time if it is unset.
   SetStartTimeIfUnset(tag_values, now);
   SetUpdateTime(tag_values, now);
+  RemoveExpired(now);
   switch (type_) {
     case Type::kDouble: {
       if (aggregation_.type() == Aggregation::Type::kSum) {
@@ -324,7 +325,14 @@ void ViewDataImpl::SetUpdateTime(const std::vector<std::string>& tag_values,
     update_times_.splice(update_times_.begin(), update_times_,
                          update_time_list_iter);
   }
+}
 
+void ViewDataImpl::RemoveExpired(absl::Time now) {
+  if (expiry_duration_ == absl::ZeroDuration() || update_times_.empty()) {
+    // No need to remove expired entries since either expiry is not set or there
+    // is no data entry yet.
+    return;
+  }
   // Remove data that has not been updated for expiry.
   while (now - update_times_.back().first > expiry_duration_) {
     const auto& tags = update_times_.back().second;
--- a/opencensus/stats/internal/view_data_impl.cc
+++ b/opencensus/stats/internal/view_data_impl.cc
@@ -65,6 +65,7 @@ ViewDataImpl::ViewDataImpl(absl::Time start_time,
       aggregation_window_(descriptor.aggregation_window_),
       type_(TypeForDescriptor(descriptor)),
       start_times_(),
+      expiry_duration_(descriptor.expiry_duration_),
       start_time_(start_time) {
   switch (type_) {
     case Type::kDouble: {
@@ -195,6 +196,7 @@ void ViewDataImpl::Merge(const std::vector<std::string>& tag_values,
                          const MeasureData& data, absl::Time now) {
   // A value is set here. Set a start time if it is unset.
   SetStartTimeIfUnset(tag_values, now);
+  SetUpdateTime(tag_values, now);
   switch (type_) {
     case Type::kDouble: {
       if (aggregation_.type() == Aggregation::Type::kSum) {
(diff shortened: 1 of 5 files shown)


bianpengyuan:
@mandarjog fyi

opencensus/stats/internal/view_data_impl.cc
```diff
@@ -302,5 +304,53 @@ ViewDataImpl::ViewDataImpl(ViewDataImpl* source, absl::Time now)
...
+    // The timeseries is tracked, update its updated time and move it to the
+    // front of the list.
+    auto update_time_list_iter = update_time_map_iter->second;
+    update_time_list_iter->first = now;
+    update_times_.splice(update_times_.begin(), update_times_,
+                         update_time_list_iter);
+  }
+
+  // Remove data that has not been updated for expiry.
+  while (now - update_times_.back().first > expiry_duration_) {
```
mandarjog, line 329: For clarity, from here down, it should be called “removeExpired”
You can call it from here. removeExpired must also be called periodically.
We don’t want the data pileup, if the next request comes much later.

bianpengyuan: This function is already triggered periodically, whenever the view has new timeseries create or any update to the existing ones. So the only case that the timeseries got created without being removed is that they are all created during an expiry duration and no update at all after that, which seems to be very rare. Considering this, do you think it is still necessary to call remove periodically?

This seems to be the cleanest way to implement, otherwise we'd either need to plumb RemoveExpired to multiple data structure in order to reuse the existing periodic merging, or create a new timer, which seems to be overkill.

mandarjog: Since we are fixing it, we should fix the known cases. I will defer to @g-easy on this.

naming: `SetUpdateTime` name gives no indication that it will purge data.

opencensus/stats/internal/view_data_impl.cc
```diff
@@ -195,6 +196,7 @@ void ViewDataImpl::Merge(const std::vector<std::string>& tag_values,
                          const MeasureData& data, absl::Time now) {
   // A value is set here. Set a start time if it is unset.
   SetStartTimeIfUnset(tag_values, now);
+  SetUpdateTime(tag_values, now);
```
mandarjog, line 199: This will read much clearer if it says
```c++
SetUpdateTime(tag_values, now);
PurgeExpired(now);
``

Comment (bianpengyuan) [reply opencensus/stats/internal/view_data_impl.cc:199]: Done.

File: opencensus/stats/internal/view_data_impl.cc [hunk 3]
```diff
@@ -302,5 +305,60 @@ ViewDataImpl::ViewDataImpl(ViewDataImpl* source, absl::Time now)
...
 }
 
+void ViewDataImpl::SetUpdateTime(const std::vector<std::string>& tag_values,
+                                 absl::Time now) {
+  if (expiry_duration_ == absl::ZeroDuration()) {
+    // No need to track last update time if expiry duration is not set.
+    return;
+  }
+
+  auto update_time_map_iter = update_time_entries_.find(tag_values);
```
g-easy, line 315: I'm worried about the performance of list::find. Why is this a list?

opencensus/stats/view_descriptor.h
```diff
@@ -82,6 +82,11 @@ class ViewDescriptor final {
   ViewDescriptor& set_description(absl::string_view description);
   const std::string& description() const { return description_; }
 
+  // Sets the desired lifetime of the view's data. If the view data is not
+  // updated at any point during this duration, it will be removed from the set
+  // of data recorded by the view when periodic aggregation happens.
+  ViewDescriptor& set_expiry_duration(const absl::Duration& expiry_duration);
```
g-easy, line 88: Please pass Duration by value, not by constref.

opencensus/stats/internal/view_data_impl.h
```diff
@@ -147,6 +155,16 @@ class ViewDataImpl {
...
 
+  using UpdateTimeList =
+      std::list<std::pair<absl::Time, std::vector<std::string>>>;
+
+  // A list of last update time for each timeseries.
+  UpdateTimeList update_times_;
+  // A map from view data tags to the last update time list iterator.
+  DataMap<UpdateTimeList::iterator> update_time_entries_;
+
+  absl::Duration expiry_duration_;
```
g-easy, line 166: const

opencensus/stats/internal/view_data_impl.h
```diff
@@ -134,6 +134,14 @@ class ViewDataImpl {
     }
   }
 
+  // If expiry duration is set, keep track of the last update time of each view
+  // data and remove data that has not been updated during the expiry duration.
```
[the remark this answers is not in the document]
bianpengyuan: Good catch! Done.

opencensus/stats/internal/view_data_impl.h
bianpengyuan: Done.

opencensus/stats/view_descriptor.h
bianpengyuan: Done.

opencensus/stats/internal/view_data_impl.cc
bianpengyuan: `update_time_entries_` is a map from tags to list entry iterator. We need the list to keep track of update time of each view data. This map is used to find the list entry to update time for in constant time.

opencensus/stats/internal/view_data_impl.cc
g-easy: I think purge on update is ok.

opencensus/stats/internal/view_data_impl.cc
g-easy: Oops, I misread it. We keep both a map and a list, so we can early exit after processing just enough of the list (L337)?

g-easy:
Please run tools/format.sh if you haven't already, and merge master branch to get the format CI fix. :)

opencensus/stats/internal/view_data_impl.cc
bianpengyuan: Yeah, the list is sorted based on last updated time, so that we don't need to scan all time series on every update.

bianpengyuan:
Hmm I dont think the test failure related to this PR:

```
//opencensus/exporters/stats/stackdriver:stackdriver_utils_test          FAILED in 0.2s
  /private/var/tmp/_bazel_travis/e4c98731ffb6c57222861c08b7d3a5cb/execroot/io_opencensus_cpp/bazel-out/darwin-fastbuild/testlogs/opencensus/exporters/stats/stackdriver/stackdriver_utils_test/test.log
//opencensus/exporters/trace/ocagent:ocagent_exporter_test               FAILED in 0.2s
  /private/var/tmp/_bazel_travis/e4c98731ffb6c57222861c08b7d3a5cb/execroot/io_opencensus_cpp/bazel-out/darwin-fastbuild/testlogs/opencensus/exporters/trace/ocagent/ocagent_exporter_test/test.log
```

Are they known flaky? Not sure how can I look at the test log file.

bianpengyuan:
Ping @g-easy any help on reviewing this again?

g-easy (approved):

PR #454, opened 2020-12-17
Disable signed/unsigned comparison warnings

KeithMoyer:
OpenCensus compares signed and unsigned values throughout the code. This appears to be a conscious decision, and one that many projects take.

It also relies on Abseil to provide a sane set of compile/warning flags. However, Abseil has recently (https://github.com/abseil/abseil-cpp/commit/5bf048b8425cc0a342e4647932de19e25ffd6ad7) stopped disabling warnings for signed/unsigned comparisons.

The base set of flags provided by Abseil no longer align with OpenCensus's approach. So, this PR adds an additional warning flag to disable these warnings directly.

### DIFF
--- a/opencensus/copts.bzl
+++ b/opencensus/copts.bzl
@@ -32,15 +32,16 @@ load(
 )
 
 WERROR = ["-Werror=return-type", "-Werror=switch"]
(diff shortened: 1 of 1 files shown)


g-easy (approved):
Thanks!

PR #466, opened 2022-03-28
[BUG FIX] Fix template syntax error

taekahn:
<N> at the end of the constructor causes a build failure in GCC 11 compiled with -std=c++20


g-easy (approved):
Thanks for the fix!

PR #469, opened 2022-09-14
Stackdriver exporter: add workload.googleapis.com as a supported external domain

dashpole:
This allows metric descriptors to be created for metrics in that domain


dashpole:
Seems like integration tests are broken.  I'd still like to contribute this, if possible

g-easy (approved):

PR #474, opened 2023-01-20
cmake improvement

wangzw:
1. support install and find_package
2. fix dependencies version
3. use external abseil and prometheus-cpp if possible
4. fix abseil build option
5. add flag to disable test
6. fix issue that target name is different if is included as sub directory
7. enable ci for cmake build


g-easy (approved):
Thanks!

